PR #23, opened 2017-02-21
MIA PT Orestes-Garcia

ogm710811:
Issues pending:
1. Display the unit price of an item with $ sign. : Apply a Regular Expression to extract the double value of unit price.
2. Clear Total Price value every time user delete or add an item.
3. Validate user input: Checking string and double values.


starter-code/css/style.css
```diff
@@ -1,77 +1,40 @@
...
-
-}
-
-.info-container {
-
-}
-
-.item {
-
+.header-content {
```
nickborbe: Great Job.  Looks great and works very well.  I noticed the issues pending list you left at the top, also good to realize what is missing even when you have a very good-looking site.

PR #24, opened 2017-02-21
MIA PT - Alejandro Jimenez

randseed1724:


starter-code/js/index.js
```diff
@@ -1,52 +1,170 @@
...
+ document.querySelector("#totalPrice").innerHTML = sum;
 }
+totalPrice();
 
-function createNewItemRow(itemName, itemUnitPrice){
 
-}
 
-function createNewItem(){
+function removeItem() {
```
nickborbe: Good job.  The page works reasonably well but your 'add more' button does not work.  
Look at the solution code I posted in the google drive to compare/contrast

PR #25, opened 2017-02-22
MIA PT Alice Guan

aliguan:


starter-code/js/index.js
```diff
@@ -1,52 +1,77 @@
...
-function getPriceByProduct(itemNode){
-
+var getProduct = document.getElementsByClassName('productName');
+var getUnitPrice = document.getElementsByClassName('productCost');
+var getUnitQty = document.getElementsByClassName('qty');
+var getTotalProductPrice = document.getElementsByClassName('totalProductPrice');
+var deleteButtons = document.getElementsByClassName('btn-delete');
+
+//Click event for delete buttons
+function deleteRow(){
```
nickborbe: only thing I would say is maybe you should change the name of this function since it's not actually deleting the row.

nickborbe:
Great solution.

PR #26, opened 2017-02-23
MIA PT Gustavo Gutierrez

ImGuss:
I ran some tests creating and deleting and checking the totals and it all seemed to work. Let me know if you find any errors. I also commented out some stuff that I didn't think was necessary but it might be so I didn't delete it.

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,4 +1,4 @@
-input {
+/*input {
   border: solid 1px black;
 }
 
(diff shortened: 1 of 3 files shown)


starter-code/css/style.css
```diff
@@ -1,4 +1,4 @@
-input {
+/*input {
```
nickborbe: Great job.  Works beautifully.

nickborbe:
Nice work.  Got everything working.

PR #27, opened 2017-02-24
MIA PT IRONHACKCART

ivan-jorge001:
EVERYTHING WORKS BUT ITERATION 5 AND I AM SORRY IS A LIL MESSY THE CODE DUE THAT I WAS TRYNNA PUT A CONSTRUCTOR SO IT GOT VERY COMPLICATED

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,163 @@
-function deleteItem(e){
 
+
+var listOfProduct = [];
+var totalsum = [];
+var total;
+var totaleverything = 0 ;
+var count = 0;
+    var quantity;
+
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,199 @@
...
 
-}
-
-function getPriceByProduct(itemNode){
-
-}
-
-function updatePriceByProduct(productPrice, index){
 
+var listOfProduct = [];
```
nickborbe: It looks like you're calculating the prices fine, but your web page won't create new items.  Check out the solution code in the Google Drive to compare/contrast

PR #28, opened 2017-02-24
MIA PT Adel Acebal

aacebal:


starter-code/js/index.js
```diff
@@ -30,21 +64,18 @@ function createItemNode(dataType, itemData){
 
 }
 
-function createNewItemRow(itemName, itemUnitPrice){
-
-}
 
 function createNewItem(){
```
nickborbe: Looks like you didn't get to this part.  Everything else looks good thought.

PR #30, opened 2017-02-25
MIA PT Sherwin Garcia

sherwino:
I know how to fix the app, but I will retire this for the sake of time.


El Carto.js
```diff
@@ -0,0 +1,182 @@
+var count = 1;
```
nickborbe: You've got everything except calculating total price and putting it on the page.  The create new item features works.

nickborbe:
Nice work.  If you have time later, go back and add the finishing touches.  It's almost all the way there.

PR #31, opened 2017-03-01
MIA PT - Juan Duque

imth3walrus:


starter-code/index.html
```diff
@@ -4,11 +4,126 @@
   <meta charset="UTF-8">
```
nickborbe: This page looks great!  
The main function works but you're missing the part that allows the user to add new items to the list and set their prices.

PR #32, opened 2017-03-01
MIA PT Dianelis Insua

dianelisinsua:


starter-code/css/style.css
```diff
@@ -1,63 +1,47 @@
-input {
-  border: solid 1px black;
-}
-
-.container {
-
+body {
```
nickborbe: Good job.  Looks like it all works properly.

PR #57, opened 2017-06-17
BCN: Miguel

Mi6u3l:


tawebbcn:
The label is not targeting the input.
Try modify the input to block type no numeric characters.
Well done !

PR #69, opened 2017-06-18
BCN-julius-suominen

jsur:


tawebbcn:
The label is not targeting the input, `for="inputID"`
Nice calc on blur.
Well done 👍

jsur:
Label change done.

PR #72, opened 2017-06-18
MAD Rosa Dominguez

rosadl:


ta-web-mad:
Congratulations, little girl! It has an impressive result. Some little errors to overcome, but nothing important.
Ole, ole, flamencou, rebujito.

PR #74, opened 2017-06-19
BCN. Carmen

carfuentes:


tawebbcn:
The`<label>` for attr must target the input ID, not the input name.

Well done ! 👍

PR #89, opened 2017-08-15
MIA Rachelle Dauphin

rachdaup93:


dsoraluz:
You did a really nice job with this exercise. You thought out a lot of the functionality and most importantly provided some nice feedback to the user each time the calculate button is pressed. The only thing I can really pick it for is wanting to see some feedback to the total price when an item that has been calculated into the total is deleted. But that is just something extra and as far as the assignment you surpassed hit all of the important things that were wanted you to concentrate on.

PR #90, opened 2017-08-15
MIA Ryan

RyanCommits:


starter-code/js/index.js
```diff
@@ -1,52 +1,85 @@
...
-  var createItemButton = document.getElementById('new-item-create');
-  var deleteButtons = document.getElementsByClassName('btn-delete');
-
-  calculatePriceButton.onclick = getTotalPrice;
-  createItemButton.onclick = createNewItem;
-
-  for(var i = 0; i<deleteButtons.length ; i++){
-    deleteButtons[i].onclick = deleteItem;
-  }
+// add new items ////////////////////
```
dsoraluz: This is a tried and true method of creating new elements and adding them to the DOM but another method you might find easier is to use string interpolation and insert a whole string rather than create variables for everything.

So instead of doing what you did below, you can try 

var newItemHtml = `<div class="itemline"><div class="item">**"+ itemName +"**</div><div class="price">**"+ price +"**</div><div><label for="quantity">QTY</label><input name="quantity" class ="quantity" type="text" value="0" /></div><div class="totalprice">$0.00</div><div><button class="btn-delete">Delete</button></div></div>`

$('.container').appendChild(newItemHtml);


This method is a little more complex because it really easy to mess the html up when you write it in one line but a good way to not mess up is just to double click the part that you want to make dynamic with a variable and once it is highlighted, press delete. After that you insert a " + variableName + " in it place.

If you need a little bit more of an explanation on this method feel free to just ask me or Nizar and we will explain it to you in a little bit of more detail.

dsoraluz:
Nice job applying flexbox to this project. Its a concept and practice that not a lot of people get or have so keep practicing where ever you can as it will make sure frontend work really stand out.

The main take away from this example is use javascript alond with the use of id's and class names to help us perform logic and give the user feedback on their interactions with our webpages and applications.

Make sure to look at the note I included in the your submitted code, I think you'll really like it.

PR #91, opened 2017-08-16
MIA - Emma

emmalinekay:
I ALMOST got it.


dsoraluz:
You did a good job in getting all of the elements and user interface in place. You also were able to figure out to get the total of each item line which is important.

PR #92, opened 2017-08-16
MIA Jubilant Jamilet

jmzelaya:
I feel like I'm close to getting the Total Price to work...


dsoraluz:
You did a good job in getting the interface and elements organized. As far as getting the total price function to work, you were definitely on the right track but you I think you may have been over thinking it.

PR #93, opened 2017-08-16
MIA Danny 

shebbbby:
I didnt use any of the functions that were there to begin with, but I think I made it work.


dsoraluz:
Good job in getting a the fundamental functionality of the calculate button and giving the user feedback as to their interactions. Some things that I would have liked to have seen is a cleaner format and layout as it was hard to track everything down on the scree.

PR #94, opened 2017-08-16
MIA - Ernesto Rodriguez

visionpoint:


dsoraluz:
Nice job using flexbox to align your items in the item lines. Getting all the practice with flexbox that you can will definitely pay off later on, especially in you find yourself in a frontend role.

PR #96, opened 2017-08-19
BCN - Mark

markvertra:
I think everything works - you can create and delete items, prices and totals calculate even with new items and you can even create new items if you delete everything!


tawebbcn:
Maybe will be a better idea make functions for deleting and creating new items. And be careful with missing semicolons!

That said, amazing job! :+1:

PR #109, opened 2017-08-21
BCN - Daily Exercise - week1/day5

MHFP:
I was able to do iteration 1-3 but got stuck with deletion and creation of items


tawebbcn:
For deleting an item, try to select the item you want to remove and its parent. Then you can use a method as .removeChild().

Anyways, good job. :+1:

PR #130, opened 2017-10-22
BCN - Dominik Lasch

dlasch089:


starter-code/index.html
```diff
@@ -9,6 +10,60 @@
...
+    <div class="product-line" id="hidden">
+      <div class="product-name">
+        <span>Ironhack Surfboard</span>
+      </div>
+      <div class="product-price price">
+        <p>$</p><span class="unit-price">25.00</span>
+      </div>
+      <div class="quantity">
+        <label>QTY</label>
+        <input type="number" class="unit-qty">
```
```
<input type="number" class="unit-qty" min="0">
```

PR #131, opened 2017-10-22
BCN - Torgeir Beyer

torgeirbeyer:


starter-code/index.html
```diff
@@ -9,6 +10,82 @@
...
+        <div class="product-name">
+          <span>Ironhack Skateboard</span>
+        </div>
+        <div class="product-cost">
+          <span class='unit-price'>50</span>
+          <p>$</p>
+        </div>
+        <div class="qty">
+          <label for="qty">QTY</label>
+          <input class='input' type="number">
```
```
<input class="number" type="number" min="0">
```

starter-code/js/index.js
```diff
@@ -1,52 +1,112 @@
...
+  // parent.appendChild(newItemRow).lastChild;
+}
 
+function createNewItem() {
+  createNewItemRow();
+  // createItemNode();
+  // createQuantityNode();
+  // createQuantityInput();
+  createDeleteButton();
+  window.onload();
```
tawebbcn: window.onload gets called when the page is loaded.

PR #139, opened 2017-10-23
almost done

cyenis:


starter-code/index.html
```diff
@@ -1,14 +1,75 @@
...
+        </div>
+        <div class="sub-total-price">
+          <span class='sub-total'></span>
+        </div>
+        <button class="btn-delete btn">Delete</button>
+      </div>
+    </div>
+    <div class="new-product">
+      <input class='new-product-name' type="text" placeholder='Product Name'>
+      <input class='new-product-price' type="text" placeholder='Product Price'>
```
tawebbcn: Better use type="number", then the number always will be a number.

starter-code/js/index.js
```diff
@@ -1,52 +1,70 @@
...
+  var products = e.currentTarget.parentNode.parentNode;
+  var deleteLine = e.currentTarget.parentNode;
+  products.removeChild(deleteLine);
+  getTotalPrice();
+}
+
+
+function createNewItemRow(itemName, itemUnitPrice) {
+  var orgProduct = document.querySelector('.product-line');
+  var newProductLine = orgProduct.cloneNode(true);
```
tawebbcn: Two things to keep in mind:
- This won't work if there are not articles remaining.
- You need to addEventListener to the new button, otherwise won't work.

starter-code/js/index.js
```diff
@@ -1,52 +1,70 @@
...
+  createDeleteButton();
+  window.onload();
+}
+
+window.onload = function() {
+  var calculatePriceButton = document.getElementById('calc-prices-button');
+  var createItemButton = document.getElementById('new-item-create');
+  var deleteButtons = document.getElementsByClassName('btn-delete');
+
+  calculatePriceButton.onclick = getTotalPrice;
```
tawebbcn: As Andre explained, is better to use addEventListener than onclick:
```
```

PR #141, opened 2017-10-23
BCN - Dafne

dafneolca:


starter-code/index.html
```diff
@@ -10,5 +10,33 @@
...
+
+  
+<!-- TESTING CODE -->
+  <div class="main">
+    <h1 class="heading-text">Ironhack<br>Merchandise Shop</h1>
+    <div class="item">
+      <span class="product">IronBubble-head</span>
+      <span class="individualPrice" type="number">25</span>
+      <span>QTY</span>
+      <input class="input" type="number">
```
```
<input type="number" min="0" />
```

starter-code/index.html
```diff
@@ -10,5 +10,33 @@
...
+      <span class="total">Total</span>
+    </div>
+    <button class="calc-prices-button" onClick="getPriceByProduct()">Calculate Prices</button>
```
tawebbcn: In this case, better use addEventListener in the JavaScript file.

PR #143, opened 2017-10-23
BCN-Felice-not done but that's as faar as I could get it 

xxxelf:


starter-code/index.html
```diff
@@ -9,6 +10,83 @@
...
+          $25.00
+        </span>
+
+      </div>
+
+      <div class=product1QTY>
+
+        <form action="/action_page.php">
+          QTY
+          <input type="number" name="QTY">
```
```
<input id="qty-input" type="number" min="0" name="quantity" value="0" />
```

PR #145, opened 2017-10-23
BCN-      Misha        cart

MythicalMichael:


starter-code/index.html
```diff
@@ -9,6 +9,81 @@
...
+    </div>
+
+    <div class="price">
+      <p>$<span class="price-product">25.00</span></p>
+
+    </div>
+    <form method="post" >
+      <div class="quantity">
+        <label for="qnt">QTY</label>
+        <input type="text" class="qnt" name="0" value="0"> </input>
```
tawebbcn: Better use type="number" for quantities.

starter-code/index.html
```diff
@@ -9,6 +9,81 @@
...
+
+
+
+
+
+  </div>
+  <footer>
+    <div class="lower-btn">
+      <div class="btn">
+        <button type="submit" class="btn-success" onclick="multiplyQntAndPrice( input , productPrice  )">
```
tawebbcn: In this case, better use addEventListener in the JavaScript file. That said, what are input and productPrice?

starter-code/js/index.js
```diff
@@ -1,3 +1,42 @@
...
+
+
+  function multiplyQntAndPrice(quantity, price){
+
+    total = quantity*price;
+
+
+
+
+   inner();
```
tawebbcn: Are you making a recursive function?

PR #146, opened 2017-10-23
BCN - Brian Marcalla

brianboyCode:


starter-code/index.html
```diff
@@ -10,5 +10,34 @@
...
+    <div class="content2">
+      <div class="product">
+        <span>ironBubble-head</span>
+      </div>
+      <div class="cost">
+        <p>$</p>
+        <span id="price1">25</span>
+      </div>
+      <div class="quantity">
+        <input id="value" type="text" name="input" value="0">
```
tawebbcn: Better use type="number" for quantities.

starter-code/js/index.js
```diff
@@ -43,8 +54,12 @@ window.onload = function(){
   var createItemButton = document.getElementById('new-item-create');
   var deleteButtons = document.getElementsByClassName('btn-delete');
 
+
+
   calculatePriceButton.onclick = getTotalPrice;
```
tawebbcn: As Andre explained, is better to use addEventListener than onclick:
```
```

starter-code/js/index.js
```diff
@@ -1,21 +1,32 @@
 function deleteItem(e){
+   console.log(e);
+   del = document.getElementsByClassName('content2');
+   divs = document.getElementsByTagName('div');
+   del.removeChild(divs);
```
tawebbcn: Be careful! If this worked, It would delete ALL the divs inside content2 (not only one).

PR #147, opened 2017-10-23
BCN - Eloi

eloigil:
Don't know at all if it's correctly updated...

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -16,6 +16,7 @@
         </div>
     </header>
     <section class= "cart">
+      <script>createNewItem()</script>
         <div id= "wrapper" class="container product wrapper">
             <script>createItemNode()</script>
             <div class= "btn delete">
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -10,5 +10,27 @@
   <title>Ironhack cart</title>
 </head>
 <body>
+    <header>
+        <div class= "container">
+            <h1>Ironhack<br>Merchandising Shop</h1>
+        </div>
+    </header>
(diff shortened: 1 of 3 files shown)


starter-code/index.html
```diff
@@ -10,5 +10,28 @@
...
+            <div class= "btn delete">
+                <span> Delete </span>
+            </div>
+
+        </div>
+    </footer >
+
+       <div id= "footer" class= "container">
+
+          <button id= "calc-prices-button" class= "btn success" onclick= 'getTotalPrice(product.price, product.quantity)' >
```
tawebbcn: `onclick= 'getTotalPrice(product.price, product.quantity)'`
Is better if you do a `addEventListener` in the js file instead declare it in the HTML.

starter-code/js/index.js
```diff
@@ -1,20 +1,48 @@
...
+function getTotalPrice(productPrice, index) {
+  product.quantity = document.getElementById("input").value;                       //value of the input
+  console.log(product.quantity);
+  index = product.quantity;
+  productPrice = product.price * index;
+  updatePriceByProduct(product.price, product.quantity);
+  var productUpdatedPrice = document.createElement('span');
+  var UpdatedPriceText = document.createTextNode(productPrice + '$');
+  productUpdatedPrice.appendChild(UpdatedPriceText);
+  document.getElementById('footer').appendChild(productUpdatedPrice);
```
tawebbcn: Doing this way, it will append a `span` element each time you click the button.
We could just take the span element and update his textContent.

PR #149, opened 2017-10-23
BCN-Desiree

dvillool:
No esta bien... step by step...


starter-code/index.html
```diff
@@ -10,5 +10,43 @@
...
+          <div class="market">
+
+              <div class="product">
+
+                  <div class="name-product"><span >Nom del producte</span>
+                  <div id="price-product" class="price-product"><span >Preu unitat del producte</span>
+
+                  <div class ="quantity-product">
+                      <label class="quantity">QTY</label>
+                      <input id="quantity-product" class="quantity" type="number" name="quantity" value="0">
```
```
<input id="quantity-product" class="quantity" type="number" min="0" name="quantity" value="0">
```

PR #151, opened 2017-10-23
BCN - Cristina Moreno

crismoreno:
Not finished the jS is not working not even for the Iteration 1. I'll be updating this one to make it work after the review so please it would be amazing if you could correct again by then. 

Thanks a lot in advance


starter-code/CSS + HTML DONE/index.html
```diff
@@ -0,0 +1,35 @@
...
+<body>
+  <div class="container">
+    <h1>Ironhack Merchandising Shop</h1>
+
+    <div class="product">
+      <span class="product-name">IronBubble-Head</span>
+      <span class="cost-of-unit">$25.00</span>
+      <div class="quantity">
+        <label for="qty">QTY</label>
+        <input id="qty-input" type="number" name="quantity" value="0" />
```
tawebbcn: Little tip, if you don't want to allow negative values, do:
```
<input id="qty-input" type="number" min="0" name="quantity" value="0" />
```
Keep in ming this is not a perfect solution.

starter-code/CSS + HTML DONE/index.js
```diff
@@ -0,0 +1,52 @@
...
+function createNewItem(){
+
+}
+
+window.onload = function(){
+  var calculatePriceButton = document.getElementById('calculate-price');
+  var quantityInput = document.getElementsByClassName('qty-input');
+  var deleteButton = document.getElementsByClassName('btn-delete');
+
+  calculatePriceButton.onclick = getTotalPrice;
```
tawebbcn: As Andre explained, is better to use addEventListener than onclick:
```
```

PR #153, opened 2017-10-23
BCN - Lucas Affini

lacazeto:


starter-code/js/index.js
```diff
@@ -1,17 +1,60 @@
...
+    }
+    else{
+      quantities.push(Number(nodeValue));
+    }
+  }
+  return quantities;
 }
 
-function updatePriceByProduct(productPrice, index){
+function updatePriceByProduct(productPrice, length){
```
tawebbcn: `length` will have the same value as `elements.length`

starter-code/js/index.js
```diff
@@ -1,17 +1,60 @@
...
 function getTotalPrice() {
+  var prices = getPriceByProduct("product-cost");
+  var quantities = getQtyByProduct("quantity");
+  var elemTotalPrice = prices.map(function(element, index){
+    return element * quantities[index];
+  });
+
+  updatePriceByProduct(elemTotalPrice, quantities.length);
 
+  var total = elemTotalPrice.reduce(function(previous, current){
```
tawebbcn: Nice! 👍

PR #156, opened 2017-10-23
BCN - Sarah Fontes

SarahFontes:
Not finished


starter-code/js/index.js
```diff
@@ -1,17 +1,35 @@
...
-
+  return itemNode.innerText;
 }
 
-function updatePriceByProduct(productPrice, index){
+function getQuantityByProduct(itemNode) {
+  return itemNode.valueAsNumber;
+}
 
+function multiplyPerQuantityAndUpdate(productPrice, quantity, index){
```
tawebbcn: We maybe can split this function in two. One for the logic of multiply and another one for the DOM element manipulation.

PR #214, opened 2018-01-16
BCN - Linda

lindaluisa:
- difficulties to approach

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,5 +1,5 @@
 function deleteItem(e){
-
+  
 }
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -35,6 +45,7 @@ function createNewItemRow(itemName, itemUnitPrice){
 }
 
 function createNewItem(){
+ var newItem = document.
```
tawebbcn: Be careful with this. When you have a syntax error all your code will fail.

PR #215, opened 2018-01-16
BCN - Jana

jjanastinn:


starter-code/js/index.js
```diff
@@ -1,23 +1,79 @@
...
+  });
+  var printResult = document.getElementById("result");
+  printResult.innerText = total;
 }
 
-function getTotalPrice() {
 
+
+function deleteItem(e){
+  e.currentTarget.parentNode.parentNode.remove();
```
tawebbcn: Maybe we can call `getTotalPrice()` for calculate the new price when we delete the item.

starter-code/js/index.js
```diff
@@ -1,23 +1,79 @@
...
+  this.price;
 }
 
-function updatePriceByProduct(productPrice, index){
 
+function getTotalPrice(product) {
+  var productCount = document.querySelectorAll(".wrapper");
+  for (var i=0; i<productCount.length; i++) {
+    var price = Number(document.getElementsByClassName("price")[i].innerText);
+    var quantity = Number(document.querySelectorAll(".quantity")[i].value);
```
tawebbcn: Try to be consistent. If you use `querySelectorAll()` use for everything. If you use `getElementsByClassName()` use for everything.

PR #218, opened 2018-01-16
BCN - Alla Roman

allarom:


starter-code/index.html
```diff
@@ -9,6 +9,30 @@
...
-<body>
+<body onload="getPriceByProduct()">
+    <div class="wrapper">
+      <h1 class="title">Ironhack Merchandising Shop</h1>
+      <div>
+        <span class="product-name">Ironhack Cup</span>
+      </div>
+      <form id="price">
+        <input value="10 Euro" type="submit">
+      </form>
```
tawebbcn: Care using an `input` inside a `form` since when you click on it has a default behaviour.

starter-code/index.html
```diff
@@ -9,6 +9,30 @@
   <script type="text/javascript" src="./js/index.js"></script>
   <title>Ironhack cart</title>
 </head>
-<body>
+<body onload="getPriceByProduct()">
```
tawebbcn: This is not a common use. 
In your js you have `window.onload = function() { ... }` that does the same as `<body onload="getPriceByProduct()">`

starter-code/js/index.js
```diff
@@ -1,13 +1,30 @@
...
+  e.getElementByClassName("quantity").value --
+}*/
 
 function getPriceByProduct(itemNode){
-
+  var x = document.getElementById("price");
+    var text = "";
+    var i;
+    for (i = 0; i < x.length ;i++) {
+        text += x.elements[i].value + "<br>";
```
tawebbcn: If you put an `id` or class to the input that you are grabbing with `x.elements[i].value` will be more easier.

PR #219, opened 2018-01-16
BCN Web Dev Jan '18 - Matt

mattmegarry:


starter-code/js/index.js
```diff
@@ -1,4 +1,92 @@
-function deleteItem(e){
+'use strict';
+
+function productTotal (){
+  console.log('Calculate prices CLICKED');
+  var pricesArray = htmlCollectionToInnerHTMLArray('unit-cost');
+  var multiplesArray = htmlCollectionToValueArray('qty-input');
+  var grandTotal = 0.00;
```
tawebbcn: This is the same as `var grandTotal = 0;`

starter-code/js/index.js
```diff
@@ -1,4 +1,92 @@
...
+    for (var j = 0; j < pricesArray.length; j++){
+      var productTotal = (pricesArray[j]) * (multiplesArray[j]);
+      document.getElementsByClassName('cost-output')[j].innerHTML = productTotal;
+      grandTotal = grandTotal + productTotal;
+    }
+  document.getElementById('grand-total').innerHTML = '$' + grandTotal; 
+  console.log(pricesArray);
+  console.log(multiplesArray);
+  console.log(grandTotal);
```
tawebbcn: Nice for debugging but you should use breakpoints instead 👌

starter-code/js/index.js
```diff
@@ -1,4 +1,92 @@
...
+  var ProdsRemaining = document.getElementsByClassName('prod-wrapper').length;
+  var newButton = '';
+  for (var i = 0; i < ProdsRemaining; i++){
+    document.getElementsByClassName('prod-wrapper')[i].id = i;
+    newButton = '<button onclick="deleteItem(\'' + i + '\')" class="btn btn-delete" type="button">Delete</button>';
+    document.getElementsByClassName('delete-box')[i].innerHTML = newButton;
+  }
+  
+  //LOL!!! I didn't read this sentence in the instructions:
+  //"Use e.currentTarget to access the "Delete" button that
```
tawebbcn: Your problem it will easier resolve with the `e.currentTarget`. 
But good job! 👍

starter-code/js/index.js
```diff
@@ -1,4 +1,92 @@
...
+  //console.log(document.getElementsByClassName('prod-wrapper')[i]);
+  document.body.removeChild(document.getElementById(id));
+  console.log('deleteItem CLICKED');
+
+  //But what if we delete a middle product? Let's make the deleteItems params go in order!
+  var ProdsRemaining = document.getElementsByClassName('prod-wrapper').length;
+  var newButton = '';
+  for (var i = 0; i < ProdsRemaining; i++){
+    document.getElementsByClassName('prod-wrapper')[i].id = i;
+    newButton = '<button onclick="deleteItem(\'' + i + '\')" class="btn btn-delete" type="button">Delete</button>';
```
tawebbcn: Try use `document.createElement()` `.setAtributte()` `.onclick =  ` instead of put it all as string.

starter-code/index.html
```diff
@@ -10,5 +10,52 @@
...
+      </div>
+      <div class="prod-info cost-output-box">
+        <span class="cost-output">0.00</span>
+      </div>
+      <div class="prod-info delete-box">
+        <button onclick="deleteItem('1')" class="btn btn-delete" type="button">Delete</button>
+      </div>          
+    </div><!--End prod-wrapper-->
+  <div class="calc-button-box">
+    <button onclick="productTotal()" id="calc-prices-button" class="btn calculate" type="button">Calculate</button>
```
tawebbcn: You don't need `onclick="productTotal()"` here since you have it in the code on line 81.

starter-code/js/index.js
```diff
@@ -1,4 +1,92 @@
...
+function htmlCollectionToValueArray(myClassName) {
+  var myHtmlColl = document.getElementsByClassName(myClassName);
+    var someArray = [];
+    var myElement = 0;
+          for (var i = 0; i < myHtmlColl.length ; i++) {
+            myElement = myHtmlColl[i].value;
+            someArray.push(parseInt(myElement));
+          }
+  return someArray;    
+}
```
tawebbcn: `htmlCollectionToValueArray` and  `htmlCollectionToInnerHTMLArray` good ones.
But as we can see, the unique difference that one use `.value` and the another one use `.innerHTML` so maybe we can do something for not repeat all the code.

PR #220, opened 2018-01-16
BCN - Pedro

pedrodiogoneto:


starter-code/js/index.js
```diff
@@ -1,52 +1,76 @@
...
+  
+  var sumPrice = 0;
+    for (var i=0; i< document.querySelectorAll(".product").length; i++) {
+      var price = document.querySelectorAll(".price")[i].innerHTML;
+      var quantity = document.querySelectorAll(".quantity")[i].value;
+      var total = price[i] * quantity[i];
+      sumPrice += total
+      document.querySelectorAll(".subTotal-price")[i].innerHTML = "$ " + total;
+    }
+ }
```
tawebbcn: For working properly, you have to convert `price` and `quantity` to `Number`.

tawebbcn: I recommend you this approach for this function since only we do the `querySelectorAll()` once:
```
function getTotalPrice() {

  var sumPrice = 0;
  var products = document.querySelectorAll(".product");
  var prices = document.querySelectorAll(".price");
  var quantities = document.querySelectorAll(".quantity");
  var subTotal = document.querySelectorAll(".subTotal-price");
  
  for (var i = 0; i < products.length; i++) {
    var price = prices[i].innerHTML;
    var quantity = quantities[i].value;
    var total = Number(price) * Number(quantity);
    subTotal[i].innerHTML = "$ " + total;
  }
}
```

PR #221, opened 2018-01-16
BCN - Rodrigo

rodrigodaoud:


starter-code/js/index.js
```diff
@@ -3,15 +3,32 @@ function deleteItem(e){
...
+    acc = acc + array[i];
+  }
+  return acc;
 }
 
 function getTotalPrice() {
-
+  var totalPrice = document.querySelectorAll('.total-price');
+    var finalPrice = foo(totalPrice);
+    document.querySelectorAll('.final-total').innerText = finalPrice;
```
tawebbcn: This returns an array, so you can't modify its innerText. Doesn't matter if you only have one element, you need to access to that first element:
```
document.querySelectorAll('.final-total')[0].innerText = finalPrice;
```
In addition, this will remove the rest of the text: "Total Price: $".

starter-code/js/index.js
```diff
@@ -3,15 +3,32 @@ function deleteItem(e){
 }
 
 function getPriceByProduct(itemNode){
-
+  var priceUnity = Number(document.querySelectorAll('.unity-price')[i].innerText);
+  return priceUnity;
 }
 
 function updatePriceByProduct(productPrice, index){
```
tawebbcn: Don't you need to call this function in getTotalPrice()?

PR #222, opened 2018-01-16
BCN - Danielle

danielleleb:


starter-code/js/index.js
```diff
@@ -35,6 +61,17 @@ function createNewItemRow(itemName, itemUnitPrice){
...
 function createNewItem(){
+//   var original = document.getElementById('repeat-this');
+//   var newItem = document.createElement(item);
+var i = 0;
+//   var clone = original.cloneNode(true);
+//   clone.id = "repeat-this" + ++i; 
+
+  // original.parentNode.appendChild(clone);
+var original = document.getElementById('repeat-this');
+var clone = original.cloneNode(true); // "deep" clone
```
tawebbcn: Tricky, but not best way to do it. 
If you clone a row that has a QTY assigned you will clone QTY's value too.
If there are no products in de cart you can't clone any product.

To improve the code doing it in this way, you can create a hidden row in the position 0 of the cart array, and you'll always copy this one :wink:

starter-code/js/index.js
```diff
@@ -35,6 +61,17 @@ function createNewItemRow(itemName, itemUnitPrice){
 }
 
 function createNewItem(){
```
tawebbcn: You need some input forms that let you define product name and price

starter-code/js/index.js
```diff
@@ -1,17 +1,43 @@
...
+  var sumPrice = 0;
+
+for (var i=0; i < document.querySelectorAll('.product-name').length; i++) {
+
 
+
+  var productCost = document.querySelectorAll('.cost')[i];
+  var productCount = document.querySelectorAll('.quantity')[i];
+  var grandTotal = parseInt(productCost.innerText) * parseInt(productCount.value);
+  var total = document.querySelectorAll('.total')[i];
```
tawebbcn: you can do this with another function like getPriceByProduct()

starter-code/js/index.js
```diff
@@ -1,17 +1,43 @@
...
 function updatePriceByProduct(productPrice, index){
 
 }
 
 function getTotalPrice() {
+  // var start = 0;
+  // var productIndex = productCost.indexOf
+  var sumPrice = 0;
+
+for (var i=0; i < document.querySelectorAll('.product-name').length; i++) {
```
tawebbcn: Instead of searching things in the document inside an iteration, you can create a variable with an array, and iterate over this array (it will make the iteration faster).

starter-code/js/index.js
```diff
@@ -1,17 +1,43 @@
...
+//   var grandTotal = parseInt(productCost[0].innerText) * parseInt(productCount[0].value);
+//   var total = document.querySelectorAll('.total');
+//   total[0].innerText = grandTotal.toFixed(2);
+// }
 
 function updatePriceByProduct(productPrice, index){
 
 }
 
 function getTotalPrice() {
```
tawebbcn: Instead of doing a very big-heavy function, break it in little functions that only do one thing.

PR #223, opened 2018-01-16
BCN - Giovanni

JoniWanKenobi:


starter-code/js/index.js
```diff
@@ -35,6 +51,15 @@ function createNewItemRow(itemName, itemUnitPrice){
 }
 
 function createNewItem(){
+  var newProdName = document.getElementById("new-prod-name").value;
+  var newProdPrice = document.getElementById("new-prod-price").value;
+  var table = document.querySelector(".product-table");
+  var row = document.querySelector(".product-row");
+  var clonedRow = row.cloneNode(true);
+  clonedRow.children[0].children[0].innerText = newProdName;
```
tawebbcn: Tricky, but not best way to do it. 
If you clone a row that has a QTY assigned you will clone QTY's value too.
If there are no products in de cart you can't clone any product.

To improve the code doing it in this way, you can create a hidden row in the position 0 of the cart array, and you'll always copy this one :wink:

PR #224, opened 2018-01-16
BCN - Filipe Neves

ncoder92:


starter-code/js/index.js
```diff
@@ -1,48 +1,129 @@
...
 function updatePriceByProduct(productPrice, index){
 
 }
 
 function getTotalPrice() {
-
+  var numOfProducts = document.querySelectorAll('.product-list-item');
+  var totalCartValue = 0;
+  for (var i = 0; i < numOfProducts.length; i++) {
+    var productPrice = Number(document.querySelectorAll('.value')[i].innerText);
```
tawebbcn: Instead of doing a very big-heavy function, break it in little functions that only do one thing.

tawebbcn: Instead of searching things in the document inside an iteration, you can create a variable with an array, and iterate over this array (it will make the iteration faster).

PR #227, opened 2018-01-16
BCN - Santi

santironhacker:


starter-code/js/index.js
```diff
@@ -2,16 +2,25 @@ function deleteItem(e){
 
 }
 
-function getPriceByProduct(itemNode){
-
+// function getPriceByProduct(nodeItem)
+function getPriceByProduct(){
+  var priceItem = document.getElementsByClassName('.price-item');
```
tawebbcn: This is not working. You have two options:
Without dot (.)
```
document.getElementsByClassName('price-item');
```
or with dot and querySelector
```
document.querySelector('.price-item');
```

starter-code/js/index.js
```diff
@@ -2,16 +2,25 @@ function deleteItem(e){
...
 
+  var quantity = document.getElementsByClassName('quantity');
+  quantity = parseInt(quantity[0].value);
+  subtotal = getPriceByProduct() * quantity;
+  return subtotal;
 }
 
 function getTotalPrice() {
-
+  return subtotal;
```
tawebbcn: subtotal is not defined in here

starter-code/js/index.js
```diff
@@ -2,16 +2,25 @@ function deleteItem(e){
...
 }
 
 function updatePriceByProduct(productPrice, index){
+  var subtotal = document.getElementsByClassName('subtotal');
+  subtotal = parseInt(subtotal[0].innerText);
 
+  var quantity = document.getElementsByClassName('quantity');
+  quantity = parseInt(quantity[0].value);
+  subtotal = getPriceByProduct() * quantity;
+  return subtotal;
```
tawebbcn: You don't need to return subtotal. You have to update the subtotal. 

It has to be somethig like:
```
document.querySelector('.subtotal').innerText = subtotal;
```

PR #228, opened 2018-01-16
En proceso

oscarjorgejorge:
lineas 43-45 no funcionan, pero no se porque..

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -11,7 +11,33 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
+  var totalCost = 0;
+  for (var x = 0; x < document.querySelectorAll('.product').length ; x ++) {
 
+  var unitValue = Number(document.querySelectorAll('.cost')[x].innerHTML);
+  var totalUnits = Number(document.querySelectorAll('input')[x].value);
+
+  var totalCostProduct = unitValue * totalUnits;
+
+  var totalCost = totalCost + totalCostProduct;
+
+
+  document.getElementsByClassName('total-cost-product')[x].innerHTML = "$" + totalCostProduct;
+  }
+
+  var heading2 = document.querySelectorAll('h2')[0];
+    if (document.querySelectorAll('h2')[0])  {
+        document.getElementById('total-price-sentence').removeChild(heading2);  
+    } 
+
+  var sentenceTotalCostProduct = document.createElement('h2');
+  sentenceTotalCostProduct.innerHTML = "Total Price: $" + totalCost;
+  document.getElementById('total-price-sentence').appendChild(sentenceTotalCostProduct);
+  document.getElementsByTagName('h2')[0].classList.add('finalprice');  
+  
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -11,7 +11,33 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
```
tawebbcn: Instead of making a super-complex function, you have some functions to be filled that will make this one lighter. Then you can call functions inside this function.

starter-code/js/index.js
```diff
@@ -11,7 +11,33 @@ function updatePriceByProduct(productPrice, index){
...
+        document.getElementById('total-price-sentence').removeChild(heading2);  
+    } 
+
+  var sentenceTotalCostProduct = document.createElement('h2');
+  sentenceTotalCostProduct.innerHTML = "Total Price: $" + totalCost;
+  document.getElementById('total-price-sentence').appendChild(sentenceTotalCostProduct);
+  document.getElementsByTagName('h2')[0].classList.add('finalprice');  
+  
+  if (typeof document.querySelectorAll('h2')[0].innerHTML.match(/[1-9]/) === "object") {
+     document.getElementById('total-price-sentence').removeChild(heading2);
```
tawebbcn: You are trying to remove a child that doesn't exist in  document.getElementById('total-price-sentence')

tawebbcn: But I don't understand why do you need this

starter-code/js/index.js
```diff
@@ -11,7 +11,33 @@ function updatePriceByProduct(productPrice, index){
...
+
+  var totalCost = totalCost + totalCostProduct;
+
+
+  document.getElementsByClassName('total-cost-product')[x].innerHTML = "$" + totalCostProduct;
+  }
+
+  var heading2 = document.querySelectorAll('h2')[0];
+    if (document.querySelectorAll('h2')[0])  {
+        document.getElementById('total-price-sentence').removeChild(heading2);
```
tawebbcn: You don't need to remove this. Instead of creating a new element (what you are doing below) you can modify the .innerText of the element that you already have

starter-code/js/index.js
```diff
@@ -11,7 +11,33 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
+  var totalCost = 0;
+  for (var x = 0; x < document.querySelectorAll('.product').length ; x ++) {
```
tawebbcn: Better create an array variable with document.querySelectorAll('.product') and iterate through the array (it is faster).

starter-code/js/index.js
```diff
@@ -11,7 +11,33 @@ function updatePriceByProduct(productPrice, index){
...
 
+  var unitValue = Number(document.querySelectorAll('.cost')[x].innerHTML);
+  var totalUnits = Number(document.querySelectorAll('input')[x].value);
```
tawebbcn: Same here, create an array outside the for and iterate using this array instead of searching in the document.

tawebbcn:
Ah, and super important. Name your pull request as BCN - Oscar

PR #229, opened 2018-01-16
BCN Stephie

kissy89:


starter-code/js/index.js
```diff
@@ -11,8 +24,14 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
+  var totalItems = Number(document.querySelectorAll(".priceforone").innerText);
 
+  var totaly = function(totalItems){
+    for(var)
```
tawebbcn: comment this out to run the code or it will break the js file

starter-code/js/index.js
```diff
@@ -11,8 +24,14 @@ function updatePriceByProduct(productPrice, index){
...
+    totaly+= totalItems[i];
+  }
 }
+  document.querySelector("#totality").innerText = "$" + totaly.toFixed(2);
```
tawebbcn: This has to be inside the function

starter-code/js/index.js
```diff
@@ -11,8 +24,14 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
+  var totalItems = Number(document.querySelectorAll(".priceforone").innerText);
```
tawebbcn: document.querySelectorAll(".priceforone") will return an array. You can not do a innerText of an array.

starter-code/js/index.js
```diff
@@ -42,11 +61,16 @@ window.onload = function(){
   var calculatePriceButton = document.getElementById('calc-prices-button');
   var createItemButton = document.getElementById('new-item-create');
   var deleteButtons = document.getElementsByClassName('btn-delete');
+  var inputs = document.querySelectorAll(".number-Items").value;
```
tawebbcn: You don't have any class named .number-Items in your HTML.

document.querySelectorAll(".number-Items") returns an array. You can't get a .value of an array.

starter-code/js/index.js
```diff
@@ -42,11 +61,16 @@ window.onload = function(){
...
 
   calculatePriceButton.onclick = getTotalPrice;
-  createItemButton.onclick = createNewItem;
+  inputs.addEventListener("input", getPriceByProduct());
```
tawebbcn:
```
inputs.addEventListener("input", getPriceByProduct);
```

PR #242, opened 2018-01-29
[MIA} Martin-Genet

CortezExpress:
i think this was definitely my most impressive.

after halfway through part two i realized i needed to organize my names of classes and ids better so if they don't math perfectly with the javascript code it's bc i was in the middle of changing them


jessicaalin:
Good job for a best try!

PR #249, opened 2018-03-23
PAR Clément

Kleemdub:
Completed, or almost. I didn't create all the functions declared in the js file. But at least it works well.
I should probably work more on abstraction.


ta-web-paris:
:1st_place_medal: Awesome!

PR #260, opened 2018-03-25
PAR Elise

elisegriset92:
It's killing me because I don't know how to make it work. 
First product is good, second one doesn't add .

Delete button doesn't work either.. 

Also I don't know which magical thing I can use in CSS to align my 2 ligns of products.. mystery ! 
thanks xx

### DIFF
--- a/food.html
+++ b/food.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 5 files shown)


ta-web-paris:
We can review the exercice tonight.
Rodrigo did a good job > https://github if you want.com/LazerLight/lab-dom-ironhack-cart

PR #262, opened 2018-03-25
PAR Brianna

ycrtw93:
Delete button works for the existing items. Don't know how to make them work for the new items created.

The total price doesn't change after an item is deleted - not sure yet how to write that into the code.

Calculate total price function works for the existing items, but don't know how to make it work for the new items created.

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,3 +1,7 @@
+* {
+  line-height: 50px;
+}
+
 input {
   border: solid 1px black;
 }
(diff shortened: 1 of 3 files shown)


ta-web-paris:
Hi Brianna,
This exercise is quite hard to do and correct as well.

In my opinion, you created too much div.
It's harder to locate the parents and child nodes after.

I found the Rodgrio's code very clean.
https://github.com/LazerLight/lab-dom-ironhack-cart

Take a look on it.
Tell us if you need more feedbacks.

PR #266, opened 2018-03-25
PAR Patrycja

kucharskap:
It's not quite done, but I honestly have no idea how to fix the rest.

For some reason I'm not able to calculate the price of any other items than the first one (I'm guessing I have to use selectorAll and then loop but I'm not sure exactly how).

Then the delete buttons works only for the existing items, not added ones.

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,3 +1,7 @@
+body{
+  text-align: center;
+}
+
 input {
   border: solid 1px black;
 }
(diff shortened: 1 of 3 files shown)


ta-web-paris:
```js
```
this line of code generates errors.

For the delete buttons, be careful to add the same class.
Otherwise, javascript can't figure out the node.

ta-web-paris:
Maggie did a great job if you want to check:

PR #268, opened 2018-03-25
PAR Bertrand

wavilo:
only the calculation button works for the first product, not the second one. the delete button doesn't work. I'm not ready to create Amazon


ta-web-paris:
Not now but later, you'll be the Amazon's nightmare.
I share to you Maggie's code.

No worries later will be easier with what you'll learn.

PR #269, opened 2018-03-26
PAR - Leo W

LeoBethel:
NOt getting access to the "Delete button" click event.


ta-web-paris:
Litlle mistake, you selected the whole div, not the button element :/
Take a look on Maggie's code:

Despite this mistake you did a good job.
This exercice is hard.

:clap:

PR #272, opened 2018-03-26
PAR Mostafa Gad

M-Gadd:


ta-web-paris:
You did a good job, it's working until the item creation (actually it create an item, but it looks like it's not related to the others)

If you want to go further, take a look at maggie's work :) 
https://github.com/ironhack-labs/lab-dom-ironhack-cart/pull/251/files

Keep up the good work !

PR #275, opened 2018-03-26
BCN - Matias & Jorge

jorgeterol:
@matuki92

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,44 +1,67 @@
-function deleteItem(e){
+function deleteItem(e) {
 
 }
 
-function getPriceByProduct(itemNode){
+function getPriceByProduct(itemNode) {
 
 }
 
-function updatePriceByProduct(productPrice, index){
+function updatePriceByProduct(productPrice, index) {
 
 }
+
+
 
 function getTotalPrice() {
+  var productArray = [document.getElementsByClassName('product')]
+  var priceProduct = document.getElementsByClassName('price');
+  var quantityProduct = document.getElementsByClassName('quantity');
+  var totalPrice = document.getElementsByClassName('total-price');
+
+  console.log(productArray)
+
+
+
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,75 @@
...
 
-function updatePriceByProduct(productPrice, index){
+function updatePriceByProduct(productPrice, index) {
 
 }
 
+
+
 function getTotalPrice() {
+  var productArray = [document.getElementsByClassName('product')]
```
tawebbcn: ```document.getElementsByClassName('product')``` already returns an array. No need to wrap it around another one

starter-code/js/index.js
```diff
@@ -1,52 +1,75 @@
...
+    var result = priceProduct[i].innerHTML * quantityProduct[i].value;
+    totalPrice[i].innerHTML = result;
+
+    totalResult += result;
+  }
+
+  document.getElementById('final-price').innerHTML = totalResult;
+
 
 }
```
tawebbcn: There's a bit too much spacing:
```
function getTotalPrice() {
  var productArray = document.getElementsByClassName('product');
  var priceProduct = document.getElementsByClassName('price');
  var quantityProduct = document.getElementsByClassName('quantity');
  var totalPrice = document.getElementsByClassName('total-price');

  var totalResult = 0;

  for (var i = 0; i <= productArray.length; i++) {
    var result = priceProduct[i].innerHTML * quantityProduct[i].value;
    totalPrice[i].innerHTML = result;
    totalResult = result;
  }

  document.getElementById('final-price').innerHTML = totalResult;
}
```

starter-code/js/index.js
tawebbcn: ```document.getElementsByClassName('product')``` already returns an array like structure. It returns an  HTMLCollection that has the .length property which you used below

PR #276, opened 2018-03-26
BCN - Lennie - Seba - Getting there! 

lenniez:
@sebachowa

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,22 +1,30 @@
+function getPriceByProduct(itemNode) {
+  var price = document.querySelector(itemNode).innerHTML;
+  return price;
+}
+
+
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,23 +1,31 @@
...
+function createQuantityInput(itemNode) {
+  var qty = document.querySelector(itemNode).value;
+  return qty;
 }
 
-function updatePriceByProduct(productPrice, index){
-
+function getTotalPrice() {
+  var price = getPriceByProduct('.price');
+  var qty = createQuantityInput('.qty');
```
tawebbcn: A note on this, it will only work for the first product on the list. If you have more products you have to use querySelectorAll

PR #277, opened 2018-03-26
BCN - Jordi & Dauv

jordicalzada:


starter-code/index.html
```diff
@@ -9,6 +9,50 @@
...
+        <input type="number" class="quantity">
+      </div>
+      <div>
+          <span>$0.00</span>
+      </div>
+      <div>
+          <button>Delete</button>
+      </div>
+      <div>
+          <button>Calculate Prices</button>
```
tawebbcn: There's an id missing here that would connect the getTotalPrice function with this one.
```var calculatePriceButton = document.getElementById('calc-prices-button');```

File: starter-code/js/index.js [hunk 2]
```diff
@@ -2,16 +2,31 @@ function deleteItem(e){
...
 
 function updatePriceByProduct(productPrice, index){
 
+  
+
 }
 
 function getTotalPrice() {
 
+  return getPriceByProduct(document.getElementsByClassName('.unit-cost')) * getQuantitybyProduct(document.getElementsByClassName('.quantity'));
```
tawebbcn: This line is too long and unreadable. Don't be afraid to break it down in to small steps. Readability is key!

starter-code/js/index.js
```diff
@@ -2,16 +2,31 @@ function deleteItem(e){
...
 function getPriceByProduct(itemNode){
 
+    var price = document.querySelector(itemNode);
+    return price;
+
+}
+
+function getQuantitybyProduct(itemNode){
+
+  var qty = document.querySelector(itemNode).innerHTML;
```
tawebbcn: This will only get the qty for the first product of the list

starter-code/js/index.js
```diff
@@ -2,16 +2,31 @@ function deleteItem(e){
 
 }
 
+
 function getPriceByProduct(itemNode){
 
+    var price = document.querySelector(itemNode);
```
tawebbcn: This will only get the price for the first product of the list

PR #278, opened 2018-03-26
BCN-Miriam-Fredrik

MiriamAparicio:
@fredrik62

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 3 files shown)


starter-code/index.html
```diff
@@ -9,6 +10,30 @@
...
+    </div>
+    <div>
+      <span id="total">$00.00</span>
+    </div>
+    <div>
+      <button class="btn-delete btn">Delete</button>
+    </div>
+  </div>
+  <div id="total-container">
+    <button onclick="getTotalPrice()" id="calc-prices-button" class="btn-success btn">Calculate Prices</button>
```
tawebbcn: The use of onclick is cheating :stuck_out_tongue_closed_eyes: In your JS there is code to help you do this without adding the onClick:
```
```

starter-code/js/index.js
```diff
@@ -2,8 +2,9 @@ function deleteItem(e){
 
 }
 
-function getPriceByProduct(itemNode){
 
+function getPriceByProduct(itemNode){
+  return itemNode[0].innerText;
```
tawebbcn: This will only return the price of the first of the products found by the document.getElementsByClassName("item-price");

starter-code/js/index.js
```diff
@@ -12,6 +13,16 @@ function updatePriceByProduct(productPrice, index){
 
 function getTotalPrice() {
 
+  var itemPriceNode = document.getElementsByClassName("item-price");
```
tawebbcn: The naming can be improved here

starter-code/js/index.js
```diff
@@ -38,13 +49,17 @@ function createNewItem(){
...
+
+
 window.onload = function(){
   var calculatePriceButton = document.getElementById('calc-prices-button');
   var createItemButton = document.getElementById('new-item-create');
   var deleteButtons = document.getElementsByClassName('btn-delete');
 
-  calculatePriceButton.onclick = getTotalPrice;
-  createItemButton.onclick = createNewItem;
+  calculatePriceButton.addEventListener("click", getTotalPrice);
```
tawebbcn: This was not needed in this case but it's a good step although it will be duplicated with your onclick

PR #279, opened 2018-03-26
BCN - Jenny - German

Fantasmy:
@germanrodrickson

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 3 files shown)


starter-code/index.html
```diff
@@ -9,6 +10,18 @@
...
+
 <body>
+  <header id="site-header"> Ironhack Merchandising Shop </header>
+    <div class="all-products">
+      <div>IronbubbleHead</div>
+      <div class="product-price">25</div>
+      <div><p>QTY</p><input class="qty" type="number"/></div>
+      <div>totalPrice</div>
+      <button onclick="deleteItem"></button>Delete</button>
+      <button onclick="calculatePrice"></button class="buttonPrice">Calculate Prices</button>
```
tawebbcn: A couple of things here, you should not have used onclick as you can use DOM Manipulation to achieve the same with:
```
```
In this case you had to add the Id to the button

PR #280, opened 2018-03-26
BCN Ioana-D Rubén-N

IoanaOrca:
@rubenaguayo

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,67 @@
-function deleteItem(e){
 
-}
 
-function getPriceByProduct(itemNode){
+// function deleteItem(e){
+//   // var button = document.getElementsByClassName('delete-button');
+//   // button.onclick = function () {
 
-}
+//   // }
+// }
 
-function updatePriceByProduct(productPrice, index){
+// function getPriceByProduct(itemNode){
 
-}
+// }
+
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,67 @@
...
+// }
+
+// function updatePriceByProduct(productPrice, index){
+
+// }
 
 function getTotalPrice() {
+  var totalPrice = document.getElementsByClassName('unit-price')[0].innerText;
+  console.log(totalPrice);
+  var quantifyOfItems= document.document.querySelector('input[placeholder]').innerText;
```
tawebbcn: For access to the value of an `input` you should use `.value` instead. But careful since you always will get a `string`

starter-code/js/index.js
```diff
@@ -1,52 +1,67 @@
...
 
-}
+// }
+
+// function updatePriceByProduct(productPrice, index){
+
+// }
 
 function getTotalPrice() {
+  var totalPrice = document.getElementsByClassName('unit-price')[0].innerText;
```
tawebbcn: A good practice is get first the element and then get whatever you need:
```
var totalPriceElement = document.getElementsByClassName('unit-price')[0];
var totalPrice = totalPriceElement.innetText
```

PR #281, opened 2018-03-26
BCN - RAPHA & SAMANTHA

raphamontenegro:
REEEAAAAALLYYYYY BLOCKED

@samanthaferreira


starter-code/js/index.js
```diff
@@ -50,3 +52,35 @@ window.onload = function(){
...
+                 
+        
+  
+
+                 
+                 
+                 
+}
+
+main()
```
tawebbcn: You don't need a main() function to be executed. The functions you declare will be executed once you trigger an event.

starter-code/js/index.js
```diff
@@ -50,3 +52,35 @@ window.onload = function(){
...
+var calculateButton = document.getElementById("calculate-btn");
+var totalAmmount = document.querySelector(".total");
+
+  var calculatePrice = function() {
+      for (var i = 0; i < quantityArray.length; i++) {
+        totalPrice+= costArray[i] * quantityArray[i];
+      }
+      // calculateButton.removeEventListener
+      console.log("monkeyBusiness");
+     totalAmmount.innerHTML = totalPrice;
```
tawebbcn: I think that what you want to do is to update all total amounts. Then what you have to do is a .querySelectorAll('.total') and update them with a for that iterates the quantity and the totalprice to assign.

tawebbcn: Once you do this you can get all the prices and calculate the sum of all prices

PR #283, opened 2018-03-26
BCN Mihai & Fabian 

fscheugenpflug:
lacking implementation!

@mihai.costantea


starter-code/js/index.js
```diff
@@ -1,17 +1,29 @@
+var quantity = document.createElement("button");
+var quantity_text = document.createTextNode("Calculate Prices");
+quantity.appendChild(quantity_text);
+document.body.appendChild(quantity);
+
+
+
+var price = document.getElementsByClassName('price');
+price.innerHTML;
```
tawebbcn: ? this line does nothing

starter-code/js/index.js
```diff
@@ -1,17 +1,29 @@
...
+var quantity = parseFloat(price[0].innerHTML) * parseFloat(document.querySelector('input').innerHTML) ;
```
tawebbcn: Be careful! You are declaring twice var quantity.

tawebbcn:
```
document.querySelector('input').value
```

PR #303, opened 2018-04-10
FTWD Miami - Ryan Martin

ryancraigmartin:


starter-code/js/index.js
```diff
@@ -1,29 +1,78 @@
...
 
 }
 
 function createQuantityInput(){
-
+// Do I use this to create a quanitity input for the first product as well as the others that follow?
 }
 
 function createDeleteButton(){
+  // QUESTION: Do I need to populate the delete button using this function on the first product as well as all the products that follow? Or do I need to get the products that are added to have a button created after the first by using a template.
```
MarcosPalaciosIII: the first delete button you create in the html, the others should generate for each item that is created.

starter-code/js/index.js
```diff
@@ -1,29 +1,78 @@
...
 
 function getTotalPrice() {
+  var totalPriceID = document.getElementById('total-price'); // Gets the totalPrice by the ID.
+  var currentQtyTotal = totalPriceID.innerHTML; // Gets the totalPrice. 
 
 }
 
 function createQuantityInput(){
-
+// Do I use this to create a quanitity input for the first product as well as the others that follow?
```
MarcosPalaciosIII: only for any product that you create will you need to create a quantity input tab for

starter-code/js/index.js
```diff
@@ -35,7 +84,12 @@ function createNewItemRow(itemName, itemUnitPrice){
 }
 
 function createNewItem(){
+  // var newItem = document.getElementsByClassName('product-wrapper');
```
MarcosPalaciosIII: basically every time you use the create button to create a new item you will go through the create function for each element needed for the product your creating.

starter-code/js/index.js
```diff
@@ -1,29 +1,78 @@
...
+  
 }
 
 function getPriceByProduct(itemNode){
+  var getPriceID = document.getElementById('cost-of-item'); // Gets the priceID
+  var itemPrice = getPriceID.innerHTML; // Gets the price from the ID.
+  itemPrice = itemPrice.replace("$",""); // Removes the $ from the price.
+  itemPrice = Number(itemPrice); // Turns price into a #.
 
+  return itemPrice;
```
MarcosPalaciosIII: for this function you can use parseInt() to convert string to integer and then multiply the price and quantity then on then return it. you can use that number to update what is needed.

MarcosPalaciosIII:
your heading in the right direction just look at the notes and remember that you can do everything you would normally do to multiply, convert, etc. with dom manipulation. Your just doing it behind the scenes. You still use functions and everything else like normal js.

PR #306, opened 2018-04-15
MIA-PTWD-Francisco Odon-Ironhack Cart

fodon11:


ta-web-mia:
It works really well! only critique would be that when you add an item it automatically starts it at a quantity of 2. Other than that, swell job!

PR #307, opened 2018-04-16
MIA-PTWD-WilliamDonayre-Ironhack Cart

williamdonayre:
Almost done


williamdonayre:
Completed "item total price" and "cart total price"
Missing;:
-delete function
-add new item using elements length +1

ta-web-mia:
nice job! -- especially with the refresh items component
missing create a new item element and delete button does not work

PR #394, opened 2018-06-19
MIA FT 061118 Alina K

akitaeva:
very much half-backed

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -9,6 +9,42 @@
   <script type="text/javascript" src="./js/index.js"></script>
   <title>Ironhack cart</title>
 </head>
+
+
 <body>
+<div>
+  <div> 
+    <h1>Ironhack <br>Merchandising Shop</h1>
+  </div>  
+
+  <div class=itemInfo>
+    <span class=itemName>IronBubble-head</span>
+    <span class="price">$25</span>
+    <form>
+        <label for="quantity">QTY</label>
+        <input type="number" min="0" class="quantity" name="orderQty" value="0">
+    </form>
+    <span class="totalPricePerItem"><!--getPriceByProduct--></span>
+    <button type="button" class="btn btn-delete">Delete</button>
+  </div>
+
+  <div class=itemInfo>
+      <span class=itemName>IronShirt</span>
+      <span class="price">$10</span>
+      <form>
+          <label for="quantity">QTY</label>
+          <input type="number" min="0" class="quantity" name="orderQty" value="0">
+      </form>
+      <span class="totalPricePerItem"><!--getPriceByProduct--></span> 
+      <button type="button" class="btn btn-delete">Delete</button>
+  </div>
+
+<div class="wrapper"> 
+<button type="button" class="btn" id="btn-success">Calculate Prices</button>  
+</div>
+
+<h2>Total Price: <span id="totalItemsPrice">$<!--getTotalPrice-->  </span></h2>
+
+</div>  
 </body>
 </html>
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,87 @@
-function deleteItem(e){
+// function deleteItem(e){
 
-}
+// }
 
 function getPriceByProduct(itemNode){
 
-}
+  var itemQuantity = document.getElementsByTagName('input')[0]; //get user's input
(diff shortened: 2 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,87 @@
-function deleteItem(e){
+// function deleteItem(e){
 
-}
+// }
 
 function getPriceByProduct(itemNode){
 
-}
+  var itemQuantity = document.getElementsByTagName('input')[0]; //get user's input
```
MarcosPalaciosIII: this is for an input tag. you should put .value at the end in order to get the value of the input. It would look like this `  var itemQuantity = document.getElementsByTagName('input')[0].value;`

starter-code/js/index.js
```diff
@@ -1,52 +1,87 @@
...
+  var itemPrice = getElementsByClassName('price')[0];  //get the item's price
```
MarcosPalaciosIII: for this one you would need to grab the info from the html tag. Since it is not an input you would use .innerHTML at the end like this . `var itemPrice = document.getElementsByClassName('price')[0].innerHTML;`

starter-code/index.html
```diff
@@ -9,6 +9,42 @@
...
+
 <body>
+<div>
+  <div> 
+    <h1>Ironhack <br>Merchandising Shop</h1>
+  </div>  
+
+  <div class=itemInfo>
+    <span class=itemName>IronBubble-head</span>
+    <span class="price">$25</span>
```
MarcosPalaciosIII: a good way to set this up in order to grab the inner html (the 25) is to set it up in a p tag then put $ then the span with the price. It would look something like this 
`
<p>
   $<span class="price">25</span>
</p>
`

starter-code/index.html
```diff
@@ -9,6 +9,42 @@
...
+        <label for="quantity">QTY</label>
+        <input type="number" min="0" class="quantity" name="orderQty" value="0">
+    </form>
+    <span class="totalPricePerItem"><!--getPriceByProduct--></span>
+    <button type="button" class="btn btn-delete">Delete</button>
+  </div>
+
+  <div class=itemInfo>
+      <span class=itemName>IronShirt</span>
+      <span class="price">$10</span>
```
MarcosPalaciosIII: same for this one. a good way to set this up in order to grab the inner html (the 10) is to set it up in a p tag then put $ then the span with the price. It would look something like this 
`
<p>
   $<span class="price">10</span>
</p>
`

starter-code/js/index.js
```diff
@@ -1,52 +1,87 @@
...
-function getTotalPrice() {
 
-}
+function getTotalPrice(itemTotals) {
 
-function createQuantityInput(){
+  getPriceByProduct(itemNode);
+  
+  console.log("calculating total cost of items in the cart now");
+  var totalItemsPrice = itemTotals.reduce(function(a, b) { return a + b; }, 0);
```
MarcosPalaciosIII: you can't get itemTotals because it is within the scope of another function

starter-code/js/index.js
```diff
@@ -1,52 +1,87 @@
...
+  var priceByProduct = itemQuantity*itemPrice;  //calculate the total for the quantity of each item 
+  console.log(priceByProduct);
 
-function updatePriceByProduct(productPrice, index){
+  var itemTotals = [];
+  itemTotals.push(priceByProduct);
 
-}
+  return priceByProduct;
+  return itemTotals;
```
MarcosPalaciosIII: you will not get to this return since this function already returns something prior to this.

starter-code/js/index.js
```diff
@@ -1,52 +1,87 @@
...
+} 
 
-function getTotalPrice() {
 
-}
+function getTotalPrice(itemTotals) {
```
MarcosPalaciosIII: when you do the on click event for the calculate price button, you are not passing anything into the argument field. this function should have an empty argument.  `getTotalPrice()`

MarcosPalaciosIII:
nice attempt on the iron cart. Just remember when working with functions that many times there can be scoping issues. If this happens you can make global variables and manipulate them from within a function.  Also note that most of the time when you do on click events for button that information wont be passed when calling a function by clicking the button. Therefore you should make on click event function with empty arguments.  

keep up the nice work.

PR #398, opened 2018-06-19
BCN Sebastian Almost done with the first iteration, I looked alot of it up on googl…

se66e:
…e, so i do need a lot of more practive

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,45 @@
-function deleteItem(e){
+var getProductPrice = function(productNode) {
+  var priceNode = productNode.querySelector(".price");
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,45 @@
...
+};
 
-}
+var calculatePrices = function(productNode) {
+  var calculateBtn = productNode.querySelector(".btn-success");
+  addEventListener.calculateBtn("click", function() {});
+};
 
-function createNewItem(){
+window.onload = function() {
```
tawebbcn: the only thing that runs in your code is what you have in here. All the functions you have defined are not called.

PR #414, opened 2018-07-03
BCN-marcposito-ES

marcposito:
Sorry I was Ill and I had no time. In the following days i'll update it and finisih it.


tawebbcn:
Okei, don't worry Marc! Take your time and once you finish do another Pull Request and we'll check it. 👍

PR #423, opened 2018-08-10
par Abi & Nerimene

nerimeneakari:
@abisummers

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -11,7 +11,26 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
+  
+  var price = document.querySelectorAll(".product-price");
+  var quatity = document.querySelectorAll(".quantity");  
+  var totalPrice = document.querySelectorAll(".total-price");
 
+
+   totalPrice.forEach(function(values, index) {
+    var total = price[index].innerHTML * quatity[index].value;
+    var parent = values.parentNode;
+   parent.innerHTML = total + " <span class= 'total-price'</span>";
+  });
+    // console.log(price)
+    // console.log(quatity)
+    // console.log(totalPrice)
+
+  var containPrices = document.querySelectorAll(".container-prices");
+  var span = containPrices.childNodes;
+  span.reduce(function(acc, cur, index) {
+   acc[index] + cur[index];
+    }, 0);
 }
 
 function createQuantityInput(){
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -11,7 +11,26 @@ function updatePriceByProduct(productPrice, index){
 }
 
 function getTotalPrice() {
-
+  
+  var price = document.querySelectorAll(".product-price");
+  var quatity = document.querySelectorAll(".quantity");
```
ta-web-paris: Defining those variables with querySelectorAll so they're arrays of values is good, so you can loop through them.

starter-code/js/index.js
```diff
@@ -11,7 +11,26 @@ function updatePriceByProduct(productPrice, index){
...
 function getTotalPrice() {
-
+  
+  var price = document.querySelectorAll(".product-price");
+  var quatity = document.querySelectorAll(".quantity");  
+  var totalPrice = document.querySelectorAll(".total-price");
+
+
+   totalPrice.forEach(function(values, index) {
+    var total = price[index].innerHTML * quatity[index].value;
```
ta-web-paris: Looping through your values to calculate the price of each line is good too, but you added an extra step that wasn't necessary. Instead of creating a "total" variable, you could have directly assigned the calculated value to the proper DOM element, like so:
```javascript
totalPrice.forEach(function(values, index) {
  totalPrice[index].innerHTML = price[index].innerHTML * quatity[index].value;
  // while looping, you directly assign the calculated value to the corresponding DOM element
});
```

Right after that, if you wanted to calculate the global price, in the same loop you could have added something like this:
```javascript
var globalPrice = document.querySelector(".all-prices");
// creates a variable holding the "global price" DOM element
var counter = 0;
// creates a counter variable to which we'll add in the loop

totalPrice.forEach(function(values, index) {
  totalPrice[index].innerHTML = price[index].innerHTML * quatity[index].value;
  // while looping, you directly assign the calculated value to the corresponding DOM element
  counter += Number( totalPrice[index].innerHTML );
  // increments the counter's value with each additional price calculated
});

  // assigns the final value of the counter to your global price
```

ta-web-paris:
This exercise is pretty confusing because you mix newly-learnt DOM manipulation techniques with pure javascript looping, but you did well!
Some small confusion about the value assignment process, but the general thinking is good :)

Good work!
Lucas

PR #424, opened 2018-08-10
PAR Pauline & Luca

lucamerzi:
@paulettepaillette


starter-code/js/index.js
```diff
@@ -1,17 +1,43 @@
...
+
+
 }
 
 function getTotalPrice() {
-
+	var totalPrices = [];
+	var prices = [ ];
+	document.querySelectorAll(".prix").forEach( el => prices.push(parseInt(el.innerHTML)) )
+
```
ta-web-paris: Pretty well done :)

ta-web-paris:
Your thinking is pretty good in the sense that you wanted to find a scalable solution, and your data manipulation is pretty good :)
Still, you complicated things a bit and right now you're doing 5 loops to calculate all this, while 1 could be enough. Like so:

```javascript
var totalPrices = document.querySelectorAll(".total");
var quantities = document.querySelectorAll("input");
var unitPrices = document.querySelectorAll(".prix");
var globalPrice = document.querySelector("h2 span");
var counter = 0;
  
for( var i=0; i<unitPrices.length; i++ ) {
  // You define the value of each total price, on each loop
  // You increment the counter with each new loop
}
// You assign the total counter value to your global price
````

Loops are "costly" in terms of computer calculation, so trying to do as little of them as possible actually improves execution times. Here there's only very few lines, but on big websites this can definitely make a difference!

Still, very good thinking for this stage of the bootcamp, keep it this way :)
Lucas

PS: very nice design haha

PR #425, opened 2018-08-10
PAR Mohamed & Emmanuelle

emmanuellejaldorau:
@SMoZam

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -6,9 +6,46 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/normalize/5.0.0/normalize.min.css">
   <link rel="stylesheet" href="./css/style.css">
-  <script type="text/javascript" src="./js/index.js"></script>
   <title>Ironhack cart</title>
 </head>
 <body>
+<h1>Ironhack Merchandising Shop</h1>
+
+<div>
+<div class= "product">
+  <div class = "product-name">IronBubble-head</div>
+  <div class= "one-unit-cost">$25.50</div>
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,76 @@
...
-}
 
-function createDeleteButton(){
+//ITERATION 1, 2 AND 3
 
-}
 
-function createQuantityNode(){
+var calculatePriceBtn = document.querySelector(".btn-success");
+var allProducts = document.querySelectorAll(".product");
```
ta-web-paris: Here you define `allProducts` as all DOM elements that have the "product" class. Which is your two products, but also the product input, and this third unwanted element makes your `totalPrice()` function go nuts :)

starter-code/index.html
```diff
@@ -6,9 +6,46 @@
...
-  <script type="text/javascript" src="./js/index.js"></script>
   <title>Ironhack cart</title>
 </head>
 <body>
+<h1>Ironhack Merchandising Shop</h1>
+
+<div>
+<div class= "product">
+  <div class = "product-name">IronBubble-head</div>
+  <div class= "one-unit-cost">$25.50</div>
```
ta-web-paris: Here, and for the quantity, you've been told to create a `<div>` with a `<span>` inside.
Adding the span would have allowed you to isolate the numeric value of the price and the quantity, saving you the pain of trying to do slices and splices later in your function. Like so:

```html
  <div> $ <span class= "one-unit-cost">25.50</span> </div>
```

ta-web-paris:
The thinking is pretty good, because aimed at finding a scalable solution, and only a little data targeting mistake prevented you from managing the global price calculation.

Continuez comme ça ! :)
Lucas

PR #443, opened 2018-08-19
MIA WDFT Sat

satga:
finally fixed the delete function & new items can also be deleted.


MarcosPalaciosIII:
Good job on the Ironhack cart lab. You seem to have an issue when creating a new item. For the Qty input it does not give the id, class or type.  The takeaway is that when adding something to html with the dom, you should test it out to assure you are appending all the information you want. 

Keep up the good work.

PR #445, opened 2018-08-20
MIA FTWD Nick Crispino

crispinonicky:


MarcosPalaciosIII:
good work on the Ironhack shopping cart lab. The delete function not working and no option to create. The main takeaway form this lesson is that manipulating the DOM is another way of being able to change the content of the html and it is not always easy. Like anything else, with a little practice you will get better at it. 

Keep up the good work.

PR #493, opened 2018-10-04
[PAR-PT] Rita & Merwan

Mrwn211:
première fonction calculate price of product

@ta-webpt-par

@mernri


starter-code/js/index.js
```diff
@@ -1,3 +1,5 @@
+var productCatalog = [];
+
 function deleteItem(e){
```
ta-webpt-par: this part can be completed with a remove() method.

starter-code/js/index.js
```diff
@@ -31,12 +33,22 @@ function createItemNode(dataType, itemData){
 }
 
 function createNewItemRow(itemName, itemUnitPrice){
```
ta-webpt-par: you'll be creating as many elements as there are inside the new row (li/div/span...)

PR #495, opened 2018-10-04
calculatePriceButton not working

arthurhoudebine:
CalculateLinePrice function not working. Seeing NaN instead of totalUnitPrice.

@ta-web-paris 
@nicolasprieur


arthurhoudebine:
Tried using class names. Still not working...
@ta-webpt-par 
@nicolasprieur

PR #504, opened 2018-10-19
Cecile and Niccolò

nicrodemus:
@gally999


nicrodemus:
we stopped at the total sum

ta-web-paris:
Good job on these iterations ! 
You can check this code if you want to see how to finish the last iterations or how to make it with the window.onload event listener : 
https://github.com/ironhack-paris-webdev-october-2018/lab-dom-ironhack-cart

(calling the first functions inside window.onload guarantees that the javascript script is not executed before the html has loaded and the DOM is ready)

PR #506, opened 2018-10-19
Antoine & Nicolas

DoneDeal0:
@Nicolasprrt


starter-code/js/index.js
```diff
@@ -1,52 +1,94 @@
...
+var globalSum = 0; 
+
+// ADD SHOPPING ITEM
+
+imgItem1.onclick = function (){
+    var firstItemName = document.createElement("li");  
+    firstItemName.innerHTML = itemNameLi1.innerHTML + ' <span>7500€</span>  <input type="text" placeholder="quantity"/> <button class="trash">🗑</button>'
+    shoppingList.appendChild(firstItemName);
+
+    globalSum+=7500;
```
ta-web-paris: It is a good thing to remember to update the global price when the user chooses a product, but hardcoding 7500 directly in the global price might lead to further problems if the user choses several products and several units of one product. 
Rather I would suggest to do things this way: 
 - add a div in your product row for the price by product
 - make a function that updates price by product and hook it to a calculate price button (or to an onchange event on your quantity input) 
- loop over the prices-by-product divs, add up their value and display the result inside tout div with class "total block"

Your code could look something like that : 

```html
  <button class="calc-price-btn">Calculate Price</button>
```
```js
//javascript:
function getTotalPrice() {
  var quantities = document.querySelectorAll(".quantity");
  var totalPrices = document.querySelectorAll(".price-by-product");
  var tempSum = 0;
  for (var i = 0; i < totalPrices.length; i++) {
    totalPrices[i].innerHTML =
      Number(unitPrices[i].innerHTML) * Number(quantities[i].value);

    tempSum += Number(totalPrices[i].innerHTML);
  }
  sumFinal.innerHTML = tempSum;
}

window.onload = function() {
  var calculatePriceButton = document.querySelector(".calc-price-btn");
};

imgItem1.onclick = function() {
  var firstItemName = document.createElement("li");
  firstItemName.innerHTML =
    itemNameLi1.innerHTML +
    ' <div><span class="unit-price">7500</span>€</div>  <input class="quantity" type="number" placeholder="quantity" value="1"/> <div>Price: <span class="price-by-product"></span></div> <button class="trash">🗑</button>';
  shoppingList.appendChild(firstItemName);

  var trashbin = firstItemName.querySelector(".trash");
  deleteLi = trashbin.parentNode;
  deleteLi.onclick = function() {
    deleteLi.remove();
  };
};
//(à refaire pour les 4 produits)

```

PR #510, opened 2018-10-20
done

nrlfrh:
@jourisque @tidochrys

Completed up until Iteration 4.


starter-code/js/index.js
```diff
@@ -1,52 +1,76 @@
...
+      }
+  };
+});
+
+//create new item (but not working)
+createBtn.onclick = function(){
+  var addProduct = document.createElement("div");
+  addProduct.setAttribute("class", ".product" );
+  addProduct.innerHTML = createProduct.value +  createProduct1.value + '<label class="quatity">Qty</label>' +' <input class="input" type="number" placeholder="0" />' + ' $<span class="total-price"></span>' + ' <button class="btn .btn-delete">Delete</button>';
+  document.body.appendChild(addProduct);
```
ta-web-paris: if you'd like your new item to be added nex to other rather than at the end of the document, you should not append it to the body element. The simplest way would be to create a wrapper div around the two div with class product, select it with querySelector, and then append your new element to it!

starter-code/js/index.js
```diff
@@ -1,52 +1,76 @@
...
+    newBtn.onclick = function() {
+        //confirm with the user that the delete is okay
+        var isOkay = confirm("Are you sure?");
+        if (isOkay) {
+          oneButton.parentNode.remove();
+        };
+    };
+    createProduct.value="";
+    createProduct1.value="";
+};
```
ta-web-paris: You are trying to mix many many steps at once in this function and a few features are not working properly : the delete function is not hooked up to the button, and also the price calculation doesn't take into account the newly created products.

If only for debugging purposes, I would advise you to separate every piece of logic inside different functions : you can extract the delete item function for exemple.  

Here, the problem is that the interpretor can't parse the html right after is has been hardcoded in innerHTML, the clean way to create elements in the dom is to use the proper functions : createElement, textContent, appendChild etc. 
Here is an exemple for extraction the creation of the delete button in a different function, but you can do the same with the label, input and span: 
```js
createBtn.onclick = function() {
  var addProduct = document.createElement("div");
  addProduct.setAttribute("class", ".product");
  addProduct.innerHTML =
    createProduct.value +
    createProduct1.value +
    '<label class="quatity">Qty</label>' +
    ' <input class="input" type="number" placeholder="0" />' +
    ' $<span class="total-price"></span>';

  document.body.appendChild(addProduct);
  var newBtn = createDeleteButton();
  addProduct.appendChild(newBtn);
  createProduct.value = "";
  createProduct1.value = "";
};

function createDeleteButton() {
  var div = document.createElement("div");
  div.className = "item-delete col-xs-3";
  var buttonNode = document.createElement("button");

  buttonNode.className = "btn btn-delete";
  buttonNode.innerHTML = "Delete";
  buttonNode.onclick = function() {
    var isOkay = confirm("Are you sure?");
    if (isOkay) {
      buttonNode.parentNode.parentNode.remove();
    }
  };
  div.appendChild(buttonNode);

  return div;
}
```
Here is a more detailed correction : https://github.com/ironhack-paris-webdev-october-2018/lab-dom-ironhack-cart/blob/master/solution-code/js/index.js

starter-code/js/index.js
```diff
@@ -1,52 +1,76 @@
...
+var createProduct = document.querySelector(".new-creation1");
+var createProduct1 = document.querySelector(".new-creation2");
+var addList = document.querySelector(".shopping-cart");
+
+calculateBtn.onclick = doCalculate;
+
+//calculate the prices + quantity
+function getTotalPrice(){
+   for ( i = 0; i < priceString.length; i++){
+    displayTotalPrice[i].innerHTML = (priceString[i].innerHTML) * (inputValue[i].value);
```
ta-web-paris: innerHTML returns a string, so you should turn these values into numbers in order for this function to work :)
```js
function getTotalPrice() {
  for (i = 0; i < priceString.length; i++) {
    displayTotalPrice[i].innerHTML =
     parseInt( priceString[i].innerHTML) * parseInt(inputValue[i].value);
  }
}
```

ta-web-paris:
You got the general concept of DOM manipulation. Try to split up every piece of logics inside different functions so that your code become easier to read and to debug, like in the code I sent you. 
Don't hesitate to come ask us if you have further questions !

PR #521, opened 2018-10-22
paris- work in progress

filippofre:


ta-web-paris:
Hey, good job on calculating the total price per item! 
Try to take some time to review the solution code : 
https://github.com/ironhack-paris-webdev-october-2018/lab-dom-ironhack-cart/blob/master/solution-code/js/index.js
With some practice, it will become easier !!

PR #527, opened 2018-10-23
Paris Heather

heyitsheather:
Accidentally saved an older version without git, but I'm looking forward to going over this with the reinforcement lesson later.


ta-web-paris:
Okay! The js file isn't there, so I can't correct, but we'll do a reinforcement on thursday, don't hesitate to ask questions and here is also the solution code if you want to review it : https://github.com/ironhack-paris-webdev-october-2018/lab-dom-ironhack-cart/blob/master/solution-code/js/index.js

PR #538, opened 2018-10-26
AMS-Lyudmyla Baron

lxs07a:


IronhackAMS:
Very nice and eloquent solution!

PR #539, opened 2018-10-27
AMS Ana Kelly

anakellyc:


IronhackAMS:
Nice work! Mind code indentation though. Besides that the code is very descriptive and clear.

PR #548, opened 2018-10-28
AMS jesco wuester

jescowuester:


IronhackAMS:
Awesome job!! Nothing to comment really. Perhaps you could think about a functionality that allows users to add their own products. Or you can have a look at media queries in order to make the page look better across different screen sizes.

PR #558, opened 2018-10-29
AMS - David Z - Finish Ironhack ShoppingCart exercise - DOM

zlobizlobi:


IronhackAMS:
Nice Zloby!  Good practice to use the window.onload event when using the DOM manipulation to make sure everything is loaded properly. Also clear code formatting, but try to implement a couple of simple comments in your js file also next time:)
Kido

PR #559, opened 2018-10-29
AMS Jude

jdhova:


IronhackAMS:
Calculating the price works very nicely! Unfortunately you can delete only one row. I think you're close to solving it though!

PR #572, opened 2018-11-01
AMS - Jonathan Chin-Sue

jce200:


IronhackAMS:
Good job Jonathan! I like the creative solution working with HTML Table elements. 
For next time, when writing code inside a body function, try to put the local variables at the top (see your addRow function)

PR #605, opened 2018-11-17
MIA-PTWD Nicolas-Castr

nicolas-castro:
- [x] Added Products

- [x] Calculated Price Per Item

- [x] Calculated Total Price


ta-web-mia:
Needs some improvements on delete function and total calculation.

PR #644, opened 2019-01-14
Paris - Francesca & Nadjie 

wowfranzik:
@NadjieVal

We struggle with the forEach loop and knowing how iterate the total price to all the products: it was confusing

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,35 @@
-function deleteItem(e){
+var button = document.querySelector(".btn-success");
+button.onclick = function() {
+  // var qtyField = document.querySelector(".qty");
+  var containerArray = document.querySelectorAll(".container");
 
-}
-
-function getPriceByProduct(itemNode){
-
-}
-
-function updatePriceByProduct(productPrice, index){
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,35 @@
...
+  });
+};
 
+function getPriceByProduct(oneContainer) {
+  console.log(oneContainer);
+  var itemQty = oneContainer.querySelector(".qty");
+  var unitCost = oneContainer.querySelector(".unit-cost");
+  var totalPrice = oneContainer.querySelector(".total-price");
+  totalPrice.innerHTML = "$" + Number(unitCost.innerHTML) * itemQty.value;
 }
```
ta-web-paris: Here is a solution code that is going a bit farther: 
```
  calculatePrice = function() {

    var counter = 0;
    
    }

  };
  
  ```

ta-web-paris:
There is going to be a reinforcement lesson on this. We can see slower all the notions and explain them again. Don't hesitate to ask us to have more explanations when needed. The part done is good though.

PR #645, opened 2019-01-14
Paris - Christophe & Delphine

Deie:
@grosbegue

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,91 @@
-function deleteItem(e){
 
-}
+var productPrice = document.querySelector(".product-cost");
+// console.log(productPrice.innerHTML);
 
-function getPriceByProduct(itemNode){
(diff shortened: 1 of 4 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,91 @@
...
+  // console.log(productTotalQty);
+  
+  var totalField = itemNode.querySelector(".item-total-price");
+      totalField.innerHTML = productQty.value * productPrice.innerHTML;
+  
+  
+    
+    totalPrice += Number(totalField.innerHTML);
+  // console.log(totalPrice);
 }
```
ta-web-paris: Well done! Here is another solution that gather everything: 
```
  calculatePrice = function() {

    var counter = 0;
    
    }

  };
  
  ```

ta-web-paris:
Well done! What you did is good. You understood how to manipulate the DOM. Great!
Continue this way!

PR #646, opened 2019-01-14
Paris Squad 107 - Kevin B. & Raja

kevbarns:
@raprocode

Done in full 💯

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,116 @@
-function deleteItem(e){
+var currency = "$";
 
+function deleteItem(e) {
+  // Selector
+  var deleteBtn = document.querySelectorAll(".btn-delete");
+
+  deleteBtn.forEach(function(e, idx) {
+    e.onclick = function() {
+      var parentBtn = e.parentNode.parentNode;
+      parentBtn.remove();
+    };
+  });
 }
 
-function getPriceByProduct(itemNode){
-
+function getPriceByProduct(itemNode) {
+  return (getPrice = document.querySelectorAll(".product-price"));
 }
 
-function updatePriceByProduct(productPrice, index){
-
-}
+function updatePriceByProduct(productPrice, index) {}
 
 function getTotalPrice() {
+  // Selector
+  var productPrice = document.querySelectorAll(".product-price-span");
+  var productPriceValue = [];
+
+  productPrice.forEach(function(e, idx) {
+    productPriceValue.push(Number(e.innerHTML));
+    console.log(productPriceValue);
+  });
+
+  // Selector
+  var productQuantity = document.querySelectorAll(".quantity-input");
+  var qty = [];
+
+  productQuantity.forEach(function(e, idx) {
+    qty.push(e.value);
+    console.log(qty);
+  });
+
+  // Selector
+  var totalPriceHolder = document.querySelectorAll(".product-total-price");
+  var totalPrice = [];
+  var totalAmount = 0;
+
+  totalPriceHolder.forEach(function(e, idx) {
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,116 @@
...
+  var deleteButtons = document.getElementsByClassName("btn-delete");
 
   calculatePriceButton.onclick = getTotalPrice;
   createItemButton.onclick = createNewItem;
 
-  for(var i = 0; i<deleteButtons.length ; i++){
+  for (var i = 0; i < deleteButtons.length; i++) {
     deleteButtons[i].onclick = deleteItem;
   }
 };
```
ta-web-paris: You put your delete button in an onclick event so that's why it's working only when you are clicking first on a delete button that is already existing. If you create a new item after having deleted the two first existing lines, it's impossible to delete it any more. 

Here is a solution to add an item and make the delete button working: 
```
  function createItem() {
    
    // I created a wrapper in the HTML, so I can appendChild the divs I'm going to create to it
    var itemsList = document.querySelector( ".items-wrapper" );

    // I create a new "div" DOM element, and save it inside my newDiv variable
    var newDiv = document.createElement( "div" );

    // I want my div to have the "wrapper" class, so I use setAttribute
    newDiv.setAttribute( "class", "wrapper" );

    // I created two inputs in the HTML, so the user can choose a name and a price for the new product
    // I gave those inputs IDs, and now I save them inside those variable to be able to retrieve their values
    var itemName = document.querySelector( "#name" );
    var itemPrice = document.querySelector( "#price" );
    
    // I want my newDiv to have the same HTML structure than the already existing items, so I create a content variable with it
    // Here I didn't use regular "" or '', but `` (backticks, or accent grave).
    // Those allow two cool things:
    //    I can break lines, and thus have a proper HTML structure with indentation: easy to write and to read
    //    I can easily use javascript variables inside it, with the ${ myVariable } syntax
    // having those variable that hold my inputs, I can easily inject them inside my content's HTML structure
    var content = `
      <div><span>${ itemName.value }</span></div>
      
      <div>$<span class="unit-price" id="unit-price">${ itemPrice.value }</span></div>
      
      <div>
        <label for="units" class="quantity">QTY</label>
        <input class="unit-quantity" id="units" placeholder="Buy ALL the things!">
      </div>
      
      <div>$<span class="total-price" id="total-price">0.00</span></div>
      
      <div>
        <button class="btn btn-delete">Delete</button>
      </div>
    `;

    // Now that I have have my content, I can simply add it inside my newDiv, using innerHTML
    newDiv.innerHTML = content;

    // My newDiv is ready, I can add it to the actual DOM using appendChild
    itemsList.appendChild( newDiv );


    // Once I'm done with creating the new item line, I clear the inputs by redefining their values to empty string ""
    itemName.value = "";
    itemPrice.value = "";
  
    // I created my new item line along with the delete button.
    // But since the delete buttons collection is created earlier, before I create my new line, my new delete button doesn't have the onclick function
    // Thus, I need to make sure it's inside the delete buttons collection by updating the deleteButtons value, like so:
    deleteButtons = document.querySelectorAll(".btn-delete");

    // And then, I need to apply the onclick function again on all the buttons of the collection,
    // so I'm sure the newly created button has the onclick function too
    deleteButtons.forEach( each => {
      each.onclick = deleteItem;
    })
  }
    
    // Here I assign my HTML button to a Javascript variable
    var createButton = document.querySelector( ".create-button" );

    // And I assign the createItem I just defined to it
    createButton.onclick = createItem;
```

ta-web-paris:
Well done, it's great! There is only a problem with the delete buttons that are activated by a function but it's something easy to solve. Continue to work hard!

PR #648, opened 2019-01-14
Paris Aurelie & Annoop

AurelieBeno:
is there an easier way to append <span> within <div> within etc.. all in 1 command ? 

@brwnrcluse

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -6,9 +6,68 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/normalize/5.0.0/normalize.min.css">
   <link rel="stylesheet" href="./css/style.css">
-  <script type="text/javascript" src="./js/index.js"></script>
+ 
   <title>Ironhack cart</title>
 </head>
 <body>
+  <div class="container">
+    <!-- <div class="All-product"> -->
+      <header>
+         <h1>Ironhack <br> Merchandising Shop</h1>
+      </header>
+  
+      <div class="checkout">
+        <div class="product"><span >IronBubble-head</span></div>
+      <div > <span class="price" > $25.00</span></div>
+      <div>  <label for="qty">QTY</label>
+        <input class ="quantity"type="number" name="qty" placeholder="0"></div>
+      <div ><span class="total" >$0.00</span> </div>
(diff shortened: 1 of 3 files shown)


starter-code/index.html
```diff
@@ -6,9 +6,68 @@
...
+    <!-- <div class="All-product"> -->
+      <header>
+         <h1>Ironhack <br> Merchandising Shop</h1>
+      </header>
+  
+      <div class="checkout">
+        <div class="product"><span >IronBubble-head</span></div>
+      <div > <span class="price" > $25.00</span></div>
+      <div>  <label for="qty">QTY</label>
+        <input class ="quantity"type="number" name="qty" placeholder="0"></div>
```
ta-web-paris: it's a good habit to add a `/` at the end of the `input tag`

starter-code/js/index.js
```diff
@@ -1,7 +1,31 @@
...
+      oneButton.onclick = function () {
+  
+        var parentContainer = oneButton.parentNode;
+  console.log(parentContainer);
+        parentContainer.parentNode.remove();
+      }
+   
+  
+  })
+
```
ta-web-paris: Here is another solution but yours is very good: 

```
  deleteItem = function( event ) {

    // event here symbolizes the click
    // event.target selects the element the click happened on, allowing us to find exactly which butotn was clicked
    // parentNode selects the current node's direct parent
    // remove() actually removes the DOM node from the DOM, deleting the line on the page
    event.target.parentNode.parentNode.remove();
  }

  // Saving the delete buttons inside a collection of buttons, so I can apply my onlick on all of them
  var deleteButtons = document.querySelectorAll(".btn-delete");

  // Looping over all the buttons of my collection
  deleteButtons.forEach( each => {

    // Applying the same onclick function on all of them, so any of them would trigger the deleteItem function when clicked
    each.onclick = deleteItem;
  });
```

starter-code/js/index.js
```diff
@@ -10,9 +34,58 @@ function updatePriceByProduct(productPrice, index){
...
+   displayPrice = sum + currentValue;
+
+   return "$" + displayPrice;
+
+})
+
+
+}
+
+
```
ta-web-paris: Here is a more condensed solution:

```
  calculatePrice = function() {

    var counter = 0;
    
    }

  };
  
  ```

PR #649, opened 2019-01-14
Paris - Christina & Alexandre

alexjoly13:
need to add the delete button.

@cibailey


ta-web-paris:
The 3rd iteration is working and is good but that would have been more handy to have it done with something that automatises the function. You seem to have understood the DOM manipulation. Continue to work hard! :)

PR #654, opened 2019-01-15
PARIS Dev Squad 107 Poeli & Jules B.

julesbadin:
almost finished


starter-code/js/index.js
```diff
@@ -1,52 +1,78 @@
...
 
-function getTotalPrice() {
+    var totalPrice = unitPrice[index].innerHTML * unitQty[index].value;
+    total[index].innerHTML = totalPrice;
+    // console.log(totalPrice);
 
-}
+    checkoutPrice = checkoutPrice + totalPrice;
+    checkoutPrice.innerHTML = checkoutPrice;
+  });
```
ta-web-paris: Here you missed a variable that is stocking the total price. It should be declared outside the forEach loop but you're almost there. 
I send you a solution that could make you understand what to do: 
```
  calculatePrice = function() {

    // Creating a counter that we'll use to calculate the big, global price
    var counter = 0;
    
    }

  };
  
  ```

ta-web-paris:
Very well done! You were very close to have a zero error lab! That's great, you understood well the DOM manipulation. It will help you a lot for your game! Continue this good work!

PR #655, opened 2019-01-15
Paris - Nadjie & Francesca

NadjieVal:
I tried another way to do it. because the last one was too complicated and not complete enough.

@wowfranzik


ta-web-paris:
Great! We are doing a reinforcement lesson tomorrow and on thursday about it so I hope we are going to go further in the different iterations. :)

PR #670, opened 2019-01-29
Fischer - Jack - FTWD - Miami

fischerjack:
Everything is done. There is a small bug when clicking the 'calculate price' button twice, but that should be fixed soon and I will add another commit.


ta-web-mia:
Goooddd Job! It is perfect! 🤩

PR #673, opened 2019-02-01
Ric Taglietti FTWD Miami

rtagl:
DOM manipulation lab

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -10,5 +10,50 @@
   <title>Ironhack cart</title>
 </head>
 <body>
+
+  <div class="header">
+    <h1>Ironhack Merchandising Shop</h1>
+  </div>
+
+
+  <div class="item">
+    <div id="itemName"><span>Iron Bubblehead</span></div>
+    <div id="bubblePrice" class="price"><span>$25</span></div>
+
+    <div class="quantity">
+      <label for="quantity">QTY: </label>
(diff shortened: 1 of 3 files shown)


ta-web-mia:
So far so good! At the moment, your cart will list all of the total prices it calculates as you recalculate the total. If you change line 39 to:

cartPrice.innerText = "Cart total is: $" + totalCartPrice;

then this will no longer be an issue. The total item and cart price are also being returned as NaN if the field is left blank. One way to avoid this issue is to set a default value of "0" within each input tag. Your price calculation and delete functions work, keep up the good work!

starter-code/index.html
```diff
@@ -10,5 +10,50 @@
...
+  </div>
+
+
+  <div class="item">
+    <div id="itemName"><span>Iron Bubblehead</span></div>
+    <div id="bubblePrice" class="price"><span>$25</span></div>
+
+    <div class="quantity">
+      <label for="quantity">QTY: </label>
+      <input name="quantity" type="number" class="quantity-input">
```
ta-web-mia: You can make the default value equal to zero so if a user doesn't add a quantity then you won't get NaN
```
<input value="0" name="quantity" type="number" class="quantity-input">
```

ta-web-mia:
Good Jobb!!! 🎊:D

PR #675, opened 2019-02-01
Henry Doce FTWD Miami

hendo12:
Delete button implemented


ta-web-mia:
Good job with the delete button. If you have extra time, try and see if you can have the cart calculate the total price for the three items you have.

PR #676, opened 2019-02-01
nunnapat.FTWD Miami

nunna19:


ta-web-mia:
The onClick for delete works, but there isn't any button or functionality for calculating price. If you have extra time, try and see if you can make the cart calculate the price for your three items.

nunna19:
I will work on that when I have a chance 
Thank you

PR #702, opened 2019-02-10
[AMS] Merle Jansen

jurlem:
Updated the chart calculations and adding new items

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,52 +1,79 @@
-function deleteItem(e){
-
-}
-
-function getPriceByProduct(itemNode){
-
-}
-
-function updatePriceByProduct(productPrice, index){
-
-}
-
-function getTotalPrice() {
-
-}
-
-function createQuantityInput(){
-
-}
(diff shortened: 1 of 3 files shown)


starter-code/js/index.js
```diff
@@ -1,52 +1,110 @@
...
 
-}
+    var calcPriceButton = document.getElementById("calculate-price-btn");
+    var total = document.getElementsByClassName("total")[0];
+    var basketValue = document.getElementById("total-sum");
+    var price = document.getElementsByClassName("price")[0];
+    var deleteButton = document.getElementsByClassName("btn-delete")[0];
+    var creatingNewBtn = document.getElementById("create");
 
-function getPriceByProduct(itemNode){
```
TA-Ironhack-Matt: There are a few lines wrong within this function. 

Please compare it to the function that i wrote from line 10 to 28 in the following repl > https://repl.it/@MattHamers2/lab-dom-ironhack-cart 

We can go over it during the next class.

PR #743, opened 2019-04-05
MIA FTWD - Nour Cherif Esoussi - April 2019

Nourcs:


ta-web-mia:
Awesome, everything works as intended and looks really good as well. My only real critique is that the createNewItemRow function isn't very organized. If you ever revisit this lab, I would consider cleaning up the code a little bit if possible. It might seem minor, but readability will become a huge part of coding once you start collaborating with other people on large projects. Other than that, everything else is fantastic. Keep it up!

PR #747, opened 2019-04-07
MIA - FTWD - APR - Nathaniel Carvajal

nathancarv:


MarcosPalaciosIII:
good job on the DOM Ironcart lab. Learning to manipulate the DOM can be very helpful when you need to make specific changes without having to completely rearrange your code. Remember when working with a partner, that if theres something you don't understand or need better explanation of to ask. The main reason many SCRUM teams are put together is so that each person can help one another with their weak points. 

Keep up the good work.

PR #751, opened 2019-04-08
MIA FTWD APR- Katy Brown

katybrown91:


ta-web-mia:
You're on the right track by selecting the innerHTML of the the cost, among other classes. It's really hard to get everything correct at first, so make sure you're console.logging every step of the way to see if you're selecting the right object. Everything in the DOM is an object, so get comfortable with dot notation and you'll end up having an easier time. And remember that you can use jQuery as well to facilitate certain methods. Keep at it!

PR #756, opened 2019-04-11
MIA FTWD ANGHY POMAR

apomarn:
Almost done.


MarcosPalaciosIII:
Good job on the Ironcart lab. I see that you did get the prices to add up for each item. Though when creating and deleting it did not work. The main takeaway from this lab is that not everything will come easily. There are some times that extra research or time will have to be used in order to reach your goal. 

Keep up the good work.

PR #773, opened 2019-06-08
LIS - Sonia #144 - Iteration 3/5

soniameller:
> Couldn't use some of the variables I declared for the  getPriceByProduct(). The work directly in the console.

> Couldn’t use the named variable $productTotal

> Completely lost after Iteration 1

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -9,6 +9,7 @@ h1 {
   font-size: 24px;
 }
 
+
 /*Added to center the heading*/
 header {
   display: flex;
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,11 +1,29 @@
+
+/* *{ 
+  border:0.3px solid #CF000F
+} */
+
+/*Added to center the heading*/
+h1 {
+  width:250px;
+  font-size: 24px;
+}
+
+/*Added to center the heading*/
+header {
+  display: flex;
+  justify-content: center;
+}
+
 input {
   border: solid 1px black;
 }
 
 .btn {
-  display: inline-block;
+  display: flex;/*changed the original to make it look like the example*/
+  justify-content: space-around; /*changed the original to make it look like the example*/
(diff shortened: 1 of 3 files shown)


GTCarmona:
Hey Sonia!

Here is what I did to use the $amountOfUnits variable inside the function getPriceByProduct()

let $amountOfUnits= document.querySelector("input"); // without ".value"
function getPriceByProduct(){
  let value= $amountOfUnits.**value***$unitPrice;
  document.querySelector(".product-price").innerHTML = value
}

What happen is, you are declaring a variable with a specific value when you say ".value". In this case it will be zero, and that's how the variable will be "saved"
So, effectively, every time you run the function, you were trying to do "0*$unitPrice"

Then way I'm doing now, and how you had it working before, every time you run the function, you FIRST access the tag "input", and then check the "value". Since it's always changing, you'll always get a different result .

I think this will help you out with the $productTotal as well.
Also, why are you using "Number" here?
function getTotalPrice() { 
  Number(document.querySelector(".total").innerHTML)+=Number(document.querySelector(".product-price").innerHTML);
}
If you console.log it, you'll see they are already numbers..
I would try without the Number , and create a variable "totalPrice" to be the outcome of that function.

One last tip, on line 17 of your index.html you have a class name "btn" witch I think it could be named "container" or something more specific, because "btn" doesn't make much sense. we usually use "btn" exclusively for buttons!

I hope that helps, if you need anything else let me know! Keep up the good work!

soniameller:
Thank you @GTCarmona !!!

**1.** 
I did what you suggested with the way of accessing the values I needed and it worked perfectly fine.

**2.**
I use the Number() method because I had strings returning in my functions so I couldn't do the math.
for example:

if I console. log --> $allProductTotals[1].innerText
return value --> "25"

if I console.log  --> Number($allProductTotals[1].innerText)
return value --> 25

I still have to continue with the next iterations, I'll pull request when I have more done.
Thanks again!

PR #784, opened 2019-06-10
Lisbon -Seb - Cart

smorel14:


GTCarmona:
Hey Seb!

Good job with the exercise, I like how you are using a lot of console.log's! It's a great practice and will always make your life easier, keep it up!

An observation, the output of your "item price" is just a number, doesn't have the dollar sign and the two decimals. This is an easy fix on the output of your function.
Also, I think you are quite able to do the "delete" button if you have the time. If you have any doubts on that let us know we'll try to help

PR #799, opened 2019-06-16
MIA-FTWD-Artem

arstrel:
Saturday code fitness


ta-web-mia:
Great work! The only super tiny thing that can be improved on is having the total price calculate and display 2 decimal places instead of just one:

image

The fact I'm nitpicking this though goes to show that there's really no issues whatsoever. Keep up the good work 👍 
- Nick

PR #803, opened 2019-06-17
FTWD-MIA Sergui Morejon

sergui20:


MarcosPalaciosIII:

You seem to be setting the .00 in the price yourself. Instead of that you can try using .toFixed() in order to get the 2 decimal places your looking for.

Keep up the excellent work.

PR #805, opened 2019-06-17
MIA-FTWD-Eduardo Gonzalez

EdoGp:


MarcosPalaciosIII:

BTW if the price is not a round number, you get a long decimal. Next time try setting your result for the final price to have ``` .toFixed(2) ``` and it will give a max of 2 decimals after any whole number.

Keep up the excellent work.

PR #806, opened 2019-06-17
MIA-FTW-Ashley Arrue

Ashleydarrue:


ta-web-mia:
So far so good! Properly calculating the price is definitely one of the hardest aspects of the lab, so definitely be proud that you managed to get that working. I'd suggest working on the delete functionality next, if you have the time. Let me know if you have any questions, and keep up the good work 👍 
- Nick

PR #807, opened 2019-06-17
MIA-FTWD-lab-dom-ironhack-cart-Pablo

PabloVazquezFayat:


MarcosPalaciosIII:

When creating each row, you seem to have the price of the product with a .toFixed() which is changing the price as it gets appended. You only need to do this on the total of the calculation between price and qty as well as when adding the total price of all items.

Keep up the excellent work.

PR #808, opened 2019-06-17
MIA-FTWD-Sergio Velikopoljski

sergiovelo17:


ta-web-mia:
Nice work! I did run into a minor bug where prices would calculate incorrectly if a number was entered into the price field without a dollar sign, but that can easily be fixed by having the input field allowing only a number, and appending a dollar sign to the front of it. Functionality is looking good though, awesome job 👍 
- Nick

PR #810, opened 2019-06-18
MIA-FTWD-Nicolle 90%

nicollegonzalez:


ta-web-mia:
the individual totals work, but no number shows up when you calculate the price of all of the items added together. Take a look into that.

ta-web-mia:
The total Price doesn't update, and I can't create my own item. This is a tricky lab, but see if you can fix those things.

PR #812, opened 2019-06-18
MIA-FTWD Taryn Chiang

tarynchiang:


ta-web-mia:
Very nicely done! Not much to critique on my end, and I commend you for doing this without the use of jQuery. Now that you're familiar with both jQuery and vanilla DOM manipulation, you're in good shape to create dynamic webpages. Keep up the good work 👍 
- Nick

PR #814, opened 2019-06-18
FTWD-MIA-MANUEL GARCIA

manuelga22:


MarcosPalaciosIII:
Great job on the ironcart lab. Working with the DOM can be tricky, but it is helpful when trying to tweak or control something in your app that normal standards cannot. Knowing how to work with the DOM also help you to understand different ways in targeting elements in the HTML which will in turn assist you in other aspects of coding.

Keep up the great work.

PR #816, opened 2019-06-19
MIA FTWD Billy Wong

billywonka3:
I completely got stuck on this one


MarcosPalaciosIII:
Good attempt on the ironcart lab. Working with the DOM can be tricky, but it is helpful when trying to tweak or control something in your app that normal standards cannot. Knowing how to work with the DOM also help you to understand different ways in targeting elements in the HTML which will in turn assist you in other aspects of coding.

Keep up the good work.

PR #817, opened 2019-06-19
MIA-FTWD - Michael Roth

mikerawth:


ta-web-mia:
Functionality is all there, nice work! Really nice work doing this with vanilla DOM manipulation. You probably would have had an easier time with jQuery, but flexing those mental muscles will do you a lot of good 💪 Keep up the good work! 
- Nick

PR #818, opened 2019-06-19
ftwd-june-2019 Gulfiya

GulfiyaIsyangulova:


ta-web-mia:
the individual totals work, but no number shows up when you calculate the price of all of the items added together. Take a look into that.

PR #819, opened 2019-06-19
MIA-FTWD-Mitch

mitchmalinin:


ta-web-mia:
Everything works except: the total calculation does not get erased after clicking the button a second time, so the total keeps getting added on to previous totals.

PR #820, opened 2019-06-19
MIA-FTWD Sams Diana 

sdiana11:


MarcosPalaciosIII:
Good job on the DOM ironcart lab. I know DOM manipulation can be tricky, but with more practice it starts to become easier. It may be something that you wont use very often, but knowing how it works and how to utilize the DOM makes it so that when you do need to use it (or read through it), you have the know how. 

Keep up the good work.

PR #821, opened 2019-06-21
MIA FTWD Matiny Louis

Matiny:


ta-web-mia:
Everything works except: when you create a new item, the price calculation doesn't accept the first number. 34 is treated as just 4, and 100 was treated as zero. If you can fix this, everything will be perfect

PR #823, opened 2019-06-26
MIA PTWD - Jesus Garcia

JesusGarcia23:
Done


ta-web-mia:
Awesome job! Just make sure you limit the data type of the new item price to a number so that it doesn't make items cost $NaN. Other than that, everything is on point. Keep it up 👍

PR #824, opened 2019-06-27
MIA PTWD Fabricio Policarpo

fpolica91:


ta-web-mia:
Nice job! I think my one suggestion is that you limit the quantity field of the items to integers rather than floats. That way you can't buy 2.5 t-shirts or some weird quantity like that. Other than that one small edge case, I can't find much to critique. Keep up the good work 👍

PR #825, opened 2019-06-28
MIA PTWD Alejandro Salas

alejotsv:
I will fix the CSS, but the functionality is there :D


ta-web-mia:
Nice work! For future reference, you can limit a user's input to a specific data type. That way you don't have to worry about them entering a string as the price (which will return NaN when you calculate the total price). If you're interested, you can learn more about it here: https://www.w3schools.com/html/html_form_input_types.asp

Keep up the good work 👍

PR #827, opened 2019-06-29
MIA PTWD - Dusty Garon

dustygaron:
Done! :)


ta-web-mia:
Nice work, the styling on this is really clean as well! I only have two small suggestions moving forward that may help:

1) Make sure that the Price of the new item is cut off at 2 decimal points. That way we can avoid having a price like $11.231231. I'd suggest using the toFixed() method.

2) Make sure the quantity is limited to integers. That way we can't buy 2.5 t-shirts.

These are only edge-cases though, and what you have is great otherwise. Keep up the good work 👍

PR #858, opened 2019-07-13
Berlin-PWDT-Dayana-not-done

dmdmdm1:
beginning

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -102,3 +102,16 @@ label {
   display: flex;
   justify-content: center;
 }
+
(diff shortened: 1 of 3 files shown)


dmdmdm1:
Here are some questions: 
- How can I look at the e.currentTarget in the console> can I select it ?
- The point of currentTarget would be to avoid loops right> so why isn't it working, why can I only delete one

ATM I CAN DELETE BUT ONLY THE FIRST OF TWO PRODUCTS

dmdmdm1:
thanks @hendrikniemann

PR #867, opened 2019-07-29
Minatsu - Homma

bmjdy151:
remove function is not working


bmjdy151:
Finished. Remove function is working.

PR #873, opened 2019-08-06
MEXPT2 - Mel

angelmc32:
Had some issues manipulating the delete buttons. I don't like the implementation that was created first, loading them into a collection and adding the event listener. I feel like it creates confusion, since the HTML collection is not my favorite type of object to work with.

It was a lot easier to just add the event listener as we created the objects, using addEventListener. It took me time to find out which node to access, and finally resolved it when creating the div that contained the button, and using newBtn.firstChildElement.addEventListener...

Great exercise tho, really enjoyed finishing it.

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,4 +1,5 @@
 * {
+  margin: 0;
   font-size: 16px;
 }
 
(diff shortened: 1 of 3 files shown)


ta-web-mex:
Excelente trabajo el punto es batallar con esto, igual debemos de encontrar la mejor manera.

PR #878, opened 2019-08-16
PAR FTWD Anaïs

anaislor:
Done !


ta-web-paris:
Good job! - maybe you could just replace the total price in the HTML each time the button is clicked :)

PR #882, opened 2019-08-19
PAR FTWD DORO

DoriDoro:
was struggeling to insert a new item to the page. the error message in the console when I press the button 'insert item' was not possible to solve and so to continue the lab


ta-web-paris:
Good job for the first iterations - come see us if you want to finish the last ones :)

PR #984, opened 2019-10-25
done

venkat-raghulan:


starter-code/js/index.js
```diff
@@ -1,11 +1,62 @@
...
+  document.querySelector("h2 span").innerHTML = total;
+}
+$calc.onclick = calcAll;
+$newprod.onclick = createNewItem;
+$tableBody.onclick = deleteProduct;
+
+function deleteProduct(evt) {
+  if (evt.target.classList.contains("btn-delete"))
+    evt.target.parentElement.parentElement.remove();
+}
```
ta-web-paris: `function deleteProduct(evt) {
  if (evt.target.classList.contains("btn-delete"))
    evt.target.parentElement.parentElement.remove();
    calcAll()
}`

You could call your calcAll function here so the price updates when you delete an element :)

ta-web-paris:
Very good :)

PR #1004, opened 2019-10-27
[PAR WDFT] Evgeny

ESR93:
Hi dears,

Please see my nooby's code =)


ta-web-paris:
Well done,
just beware of 
` let addRow = $productList[0].cloneNode(true)`
If you delete all the products, there are no nodes to clone, so you cannot create any products anymore :)
Everything working apart from that, well done.

PR #1011, opened 2019-10-28
[PAR-FTWD]-LOUISE

LouiseMarissal:
"almost done, can't update price and quantity with the newItem instead of it has the same value of the previous div"


starter-code/js/index.js
```diff
@@ -1,52 +1,102 @@
...
-
-function createNewItemRow(itemName, itemUnitPrice){
-
+// call the function
+deleteItem();
+
+// print the price for each item
+function getPriceByProduct() {
+  containers.forEach(function(container) {
+    var productPrice = container.querySelector(".product-price").textContent;
```
ta-web-paris: The reason why the listeners on the delete button / updating the price of each elements aren't working is because you are referencing to the same array when the pages loads,
Here, the "containers" you are referencing are the first containers that are present on the page, so the node list only has 2 elements at that point.
What you need to do is just query selector once again the elements once you calculate the price, so you only to add this at the start of your function.
` var containers = document.querySelectorAll(".div-container");`

starter-code/js/index.js
```diff
@@ -1,52 +1,102 @@
...
 }
-
-function createItemNode(dataType, itemData){
-
+// delete Item from the button
+function deleteItem() {
+  containers.forEach(function(container) {
+    var deleteButtons = container.querySelector(".btn-delete");
+    deleteButtons.addEventListener("click", function() {
+      container.remove(container);
```
ta-web-paris: same for the getPriceByProduct,
add this at the start of your function:
` var containers = document.querySelectorAll(".div-container");`

ta-web-paris:
Little errors, but it's good overall : )

PR #1019, opened 2019-10-28
MIA-ftwd-oct2019 Gil (gmsy)

gmsy:
First commit for that repo, iterations 1.1 and 1.2 are completed.


ta-web-mia:
So far so good! We won't end up using DOM Manipulation much moving forward, but hopefully you found the JavaScript practice to be beneficial :)
- Nick

PR #1020, opened 2019-10-28
LIS-FTWV-MariaLoureiro-174

mloureiro93:
@ironhackLis-webTeam I haven't managed to make it work even after I compared it with the class solutions :(


GTCarmona:
hey Maria! 

I checked your code, your first function is perfect, the problem was the way your were looping on the calculateAll function() I changed a little bit and left some console.logs for you to visualize, take a look and let me know if you have any questions:
```
function calculateAll() {
  let $total = 0;
  let table = document.querySelectorAll(".product");
  console.log(table)
  console.log(typeof table)
  // in this scenario the regular for loop would not work, because "table" is an HTML collection, and the information we want is inside an object, so it does not have a length
  // So, here we are sayin, for evey product class inside the table variable, we are calling the function updateSubtotal
updateSubtotal
  for (let $product of table) {
    
    $total+= updateSubtotal($product)
   }
   console.log($total)
  
}
```

Keep up the hard work and call us if you need anything (:

josecarneiro:
This branch was erroneously open from my branch to the main branch, not Maria's branch to my branch. As such, I have to close it in order to do new pull requests. Cheers!

PR #1021, opened 2019-10-29
MIA FTWD Oct 2019 - Julian Phillips

Kadus90:


ta-web-mia:
Nice job! Although we won't be using DOM Manipulation too much moving forward, hopefully you found the practice to be beneficial. Keep up the great work :)
- Nick

PR #1025, opened 2019-10-29
MIA FTWD OCT - Avrahm Kleinholz

avrahm:
…and create a new product


ta-web-mia:
Nice work! Sorry for the delay, but just wanted to get back to you and let you know that all the functionality checks out. Although we won't be using DOM manipulation very much moving forward, it's very important in developing your JavaScript skills. Keep it up :)
- Nick

PR #1026, opened 2019-10-30
MIA FTWD Masaki Endo

mendosaurus:


mendosaurus:
In Iterration 5,
 // why do you need to add class only to <tr>?
  newRow.classList.add('product')
  newRow.innerHTML = newRowInnerHtml
  newRow.querySelector('.btn-delete').onclick = deleteTheRow
  document.querySelector('tbody').appendChild(newRow)

  // set the input value empty afterwards
  document.querySelector('.new > td > input[type="text"]').value = ""
  document.querySelector('.new > td > input[type="number"]').value = ""

ta-web-mia:
Hey Masaki! Sorry for the delay, I just now had a chance to catch up on these Pull Requests.

And to answer your question, you don't necessarily need to add the .product class to only the <tr> tag, but we want to minimize the number of classes we have to add. When you're working on a website, you might not necessarily have access to changing the existing HTML code. In that sense, it's best to minimize the use of classes and to use mainly query selectors instead. Hopefully that makes sense!
- Nick

PR #1038, opened 2019-11-07
[MX-PTWD-102019] - Tonalli R. & Gustavo C.

t3rodrig:


ta-web-mex:
the cart is not working :(

PR #1065, opened 2020-01-14
[BER-FT-012020] Jacqueline C. & Nora N.

chenjacqueline:


starter-code/js/index.js
```diff
@@ -1,11 +1,76 @@
...
+
+function deleteProduct(event) {
+  console.log("Delete things");
+  const currentButton = event.target;
+  const currentProduct = currentButton.parentNode.parentNode;
+  currentProduct.parentNode.removeChild(currentProduct);
+}
+
+// 1. Assign an event listener to the create btn
+$create.onclick = createProduct;
```
DalinaLeila: I would put all the eventListeners at the top just to make the code cleaner :)

PR #1066, opened 2020-01-14
[BER-FT]Marcel Jason

jfrank325:


starter-code/js/index.js
```diff
@@ -1,11 +1,55 @@
...
+  document.querySelectorAll('.product').forEach(function(product) {
+    // console.log(product);
+    grandTotal += updateSubtot(product);
+  });
+  // display the grand total for all the selected products:
+  document.querySelector('h2 > span').innerText = grandTotal;
+  // console.log(grandTotal);
+}
+// click event connected to the function to calculate everything:
+$calc.onclick = calcAll;
```
DalinaLeila: I would put all the onclick at the top :)

starter-code/js/index.js
```diff
@@ -1,11 +1,55 @@
...
+
+function deleteProduct() {
+  const entireProduct = document.querySelector('.product');
```
DalinaLeila: work with evenTarget here. At the moment your function is only working because you put the class .product only to the first element of the table. 
The delete button at the moment does not work for the other products. 

EventTarget will query exactly the product button you clicked on. Using parentNode you can then select the parent element and delete the entire row using removeChild

starter-code/js/index.js
```diff
@@ -1,11 +1,55 @@
...
+  const allProducts = document.querySelector('#cart tbody');
+  allProducts.removeChild(entireProduct);
 }
-$calc.onclick = calcAll;
\ No newline at end of file
+$delete.onclick = deleteProduct;
+
+function addProduct() {
+  const allProducts = document.querySelector('#cart tbody');
```
DalinaLeila: addProduct also does not work at the moment. Here you would need to replicate how a product row looks (using the **same** classes. You only need to replace the innerHTML for 

```
 <span>IronBubble-head</span>
 <td class="pu">$<span>25.00</span></td>
```


using the user input (`.value`)

PR #1067, opened 2020-01-14
[BER0120-FT] Vaso-Kiyohide

vasilikibalidou:


starter-code/js/index.js
```diff
@@ -1,11 +1,55 @@
...
+    total += updateSubtot(product);
+  });
+  document.querySelector("h2>span").innerHTML = total;
 }
-$calc.onclick = calcAll;
\ No newline at end of file
+$calc.onclick = calcAll;
+
+$cart.querySelectorAll(".product>.rm>.btn-delete").forEach(function(button) {
+  button.onclick = function(event) {
```
DalinaLeila: this works but only for elements that are not created by the user. You still need to call this as well when you create the new product

starter-code/js/index.js
```diff
@@ -1,11 +1,55 @@
...
+    $cart.removeChild(event.target.parentElement.parentElement);
+  };
+});
+
+document.querySelector("#create").onclick = function(event) {
```
DalinaLeila: 💯 💯

PR #1068, opened 2020-01-14
[BER - FT] Mia - Stjepan

stifamajstor:
missing add item

### DIFF
--- a/starter-code/js/index.js
+++ b/starter-code/js/index.js
@@ -1,11 +1,32 @@
-var $cart = document.querySelector('#cart tbody');
-var $calc = document.getElementById('calc');
+var $cart = document.querySelector("#cart tbody");
+var $calc = document.getElementById("calc");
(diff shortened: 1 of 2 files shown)


starter-code/js/index.js
```diff
@@ -1,11 +1,32 @@
...
\ No newline at end of file
+
+$calc.onclick = calcAll;
+
+function deleteOnClick(element) {
+  element.querySelector(".rm > button").onclick = function() {
+    $cart.removeChild(element);
+  };
+}
+document.querySelectorAll("#cart > tbody > tr").forEach(ele => {
```
fenrew: This right here will create an error if you would ever do the "add new product" task. The reason for this is that this line right here will never be updated => To solve this, I would have put line 30-32 within your deleteOnClick function and then you can just call this deleteOnClick whenever you add a new product.

PR #1069, opened 2020-01-15
BE_FT_012020_Shahinya_Akua

akups:


starter-code/js/index.js
```diff
@@ -1,11 +1,60 @@
...
+function calcAll(price) {
   // Iteration 1.2
+  let totalPrice = 0;
+
+  document.querySelectorAll(".product").forEach(function(row) {
+    totalPrice += updateSubtot(row);
+  });
+
+  console.log(totalPrice);
+  let button = document.querySelector("h2>span");
```
fenrew: The variable name "button" is a bit misleading here as this is not a "button" but just plan text showing the total price. Making good variable names is super important when you start working in a company with large projects

starter-code/js/index.js
```diff
@@ -1,11 +1,60 @@
...
+  button.innerText = totalPrice;
+  // document.querySelector("h2>span").innerText = totalPrice;
 }
-$calc.onclick = calcAll;
\ No newline at end of file
+$calc.onclick = calcAll;
+//product.querySelector(".subtot span").innerHTML = subTotal;
+
+/*function addRow(tableID) {
```
fenrew: If you want to add a new product to the cart I believe the easier way is to:
// Get the create button and say: createButton.onclick ==> and within this onclick function you can get the different input values (name of product, price of product, etc) and just create a new element to the tbody. Below is an example code as it is a bit difficult to explain in a comment, and as always, feel free to ask me in person if it is unclear:

let oldEle = document.querySelector("#cart > tbody > tr:nth-child(1)") 
const newEle = oldEle.cloneNode(true)  // These two lines just copies the format of the shopping cart, so i don't have //to remake it

function createNewProduct() {
  let createDiv = document.querySelector("#cart > tfoot > tr") // This is the "create new product div"
  let name = createDiv.querySelector("td:nth-child(1) > input").value // This is the name input for the new product
  let priceUnit = createDiv.querySelector("td:nth-child(2) > input").value // This is the price per unit input for the new //product
  let newTrElement = newEle.cloneNode(true)  // this create a copy of the <tr> product element

  newTrElement.querySelector("td.name > span").innerHTML = name // sets the name of the new product
  newTrElement.querySelector("td.pu > span").innerHTML = priceUnit // sets the price per unit of the new product
  addDeleteOnClick(newTrElement) // adds the onclick for the delete button (if i want to remove the product again)

  document.querySelector("#cart > tbody").appendChild(newTrElement) // adds the element to the carts tbody
}

PR #1080, opened 2020-01-16
[BER-FT] - Konstantin-Ozhan

ozhann:
Done.


starter-code/js/index.js
```diff
@@ -1,11 +1,52 @@
...
+  });
+
+  document.querySelector("body > h2 > span").innerHTML = totalValue;
+}
+
+function deleteProduct(event) {
+  let currentEventTarget = event.currentTarget; //the delete button you clıcked on
+
+  if (currentEventTarget.parentNode) {
+    currentEventTarget.parentNode.parentNode.parentNode.removeChild(
```
fenrew: This seems a bit messy as it can be difficult for someone to understand what you are actually removing. 
I think it would be cleaner to write something like:
document.querySelector("#cart > tbody").removeChild(currentEventTarget.parentNode.parentNode)
However it is still unclear what is actually being removed here.

Below is another way of solving the delete button challenge that may be a bit more clear to any person in the future to what is going on in the code. If this way or your way is the best is something that can always be discussed:)

function addDeleteOnClick(element) {
  element.querySelector(".rm > button").onclick = function(){
    document.querySelector("#cart > tbody").removeChild(element)
  }
}

document.querySelectorAll("#cart > tbody > tr").forEach((ele) => {
  addDeleteOnClick(ele)
})

starter-code/js/index.js
```diff
@@ -1,11 +1,52 @@
...
+      currentEventTarget.parentNode.parentNode
+    );
+  }
 }
-$calc.onclick = calcAll;
\ No newline at end of file
+
+function addRow() {
```
fenrew: This seems to be half done. One way to solve this challenge is like this:

let oldEle = document.querySelector("#cart > tbody > tr:nth-child(1)") 
const newEle = oldEle.cloneNode(true)  // These two lines just copies the format of the shopping cart, so i don't have //to remake it

function createNewProduct() {
  let createDiv = document.querySelector("#cart > tfoot > tr") // This is the "create new product div"
  let name = createDiv.querySelector("td:nth-child(1) > input").value // This is the name input for the new product
  let priceUnit = createDiv.querySelector("td:nth-child(2) > input").value // This is the price per unit input for the new //product
  let newTrElement = newEle.cloneNode(true)  // this create a copy of the <tr> product element

  newTrElement.querySelector("td.name > span").innerHTML = name // sets the name of the new product
  newTrElement.querySelector("td.pu > span").innerHTML = priceUnit // sets the price per unit of the new product
  addDeleteOnClick(newTrElement) // adds the onclick for the delete button (if i want to remove the product again)

  document.querySelector("#cart > tbody").appendChild(newTrElement) // adds the element to the carts tbody
}

PR #1082, opened 2020-01-16
[PAR - WDFT] - Anthony B. [ongoing]

abrebion:


starter-code/js/index.js
```diff
@@ -1,11 +1,77 @@
...
+function display2Decimals(n) {
+  return Number(n).toFixed(2);
+}
+
+$calc.onclick = calcAll;
+createBtn.onclick = appendProduct;
+
+(function() {
+  calcAll();
+})();
```
frankmarmier: You don't need to use an IIFE over here since you are only executing a function.

` calcAll();  `

frankmarmier:
There isn't really much to correct on this lab, smart solution, nice touch on handling the update on the inputs aswell.
Good job : )

PR #1084, opened 2020-01-16
[PAR-WDFT-0120] - Yannick Bourenane - ongoing

yannick-bourenane:
It works, but :
- I could loop the function updateSubtot in my calcAll foreach instead of looping a second time in it
- I duplicate some code for the delete button, need to use it only once

on my todolist

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -1,10 +1,12 @@
 <!DOCTYPE html>
 <html lang="en">
+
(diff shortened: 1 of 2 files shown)


Jepassaispar:
It's really good ! I guess you already know how to make those 2 optimisations but I did them on my side so tell me if you want a hint :)

PR #1085, opened 2020-01-16
[PAR-WDFT-0120] - FRANCK.TREBOIT.ironhack-cart - ongoing

franck-treboit:
Can you please analyze this ? thanks !!!

### DIFF
--- a/starter-code/css/style.css
+++ b/starter-code/css/style.css
@@ -1,18 +1,21 @@
-table#cart {
+table#cart,  table#addCart  {
   width:100%;
   border-collapse:collapse;
 }
 
-table#cart th,
+table#cart th,  table#addCart th
 table#cart td {
   padding:.5em;
 }
 
-table#cart td {border:1px solid;}
+table#cart td ,   table#addCart td {border:1px solid;}
 
-table#cart tfoot {
+table#cart tfoot,   table#addCart toot {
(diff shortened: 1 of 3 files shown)


Jepassaispar:
Hey Franck, first of all I really like the name of your identifyAndDestroy function 💣 

Apart of that few feedbacks for you 👍 

- The event listener onclick that you are doing on the window is a bit unsafe, it would 99% of the time be better to use dom selector to select specifically the element where you want to do an event instead of doing the event on the whole window and select the action depending on what part you are clicking on. That is because first when you are clicking in the html document (not on an element inside), you will have an error in the console, that is something you want to avoid, second is if you want to do another click event on another element of this page, you will have conflict with the window click event.
- I think it is cool that you did an onchange event on the subtot price :) instead of doing it when you click on the calculate price.
- Some of the name of your variables could be clearer, that's totally subjective but it took me a bit of time to understand the concept behind the names (ex:` var createVV = document.getElementById("createValueValue");` is confusing)
- You have also on your create row some html elements that are dispensable (2 `td` that have nothing in it) and it is just this row that needs some css but the rest is good imo :)

You are almost there, you can improve a bit, I will close the pull request in the remaining time, if you want us to review it again, you can reopen the pull request and comment it.

PR #1088, opened 2020-01-16
[PAR-WDFT] Camille - ongoing

camilledlr:


Jepassaispar:
Super, very good that you used cloneNode to create a new product, bravo :)

Jepassaispar:
Just the delete button is not working on the new products, a way of make it work could have been to put your deletebutton selector in a function, call it onload and whenever you create a product

PR #1090, opened 2020-01-17
[PAR-WDFT] - Nina - ongoing

Gorlanova:
Code is a little bit redundant, but if I didn't declare var items in each function it wouldn't work... (I had placed it at the top with the others but it never refreshed value of items when I created new products)

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -3,6 +3,8 @@
 <head>
   <meta charset="UTF-8">
   <link rel="stylesheet" href="./css/style.css">
+  <script type="module" src="js/index.js"></script>
+
   <title>Ironhack cart</title>
 </head>
 <body>
@@ -18,6 +20,7 @@
         <th>Action</th>
       </tr>
     </thead>
+    
     <tbody>
       <tr class="product">
         <td class="name">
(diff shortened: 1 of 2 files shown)


Jepassaispar:
There is no issue if you redeclare variable inside functions, the idea is if you declare them in the global scope, use them, then if you redeclare in a function, use them again.
The issue would be to decalre a var and not use it like tyour `var $delete`.
For exemple it is normal to declare your var $items in both functions createNew and checkDelete because each time you will have to get the new array of items elements.
So you also won't have to pass $items as a paramater to your checkDelete function.
One other thing I noticed is when you delete an element after creating one, your delete buttons will not work anymore because you are erasing everything in the element with       `item.innerHTML = "";`
 but the element still exists so your foreach in the `var $items = $cart.querySelectorAll(".product");` will iterate in an empty element and so the loop will break. A way to solve that would be to do : `item.parentNode.removeChild(item);`
If something is not clear, feel free to check with me :)

Gorlanova:
Ah yes that makes sense ! 
I’ll rework all the labs before Monday so I’m up to date with the feedback for Front-End module :)

Thank you !!

PR #1091, opened 2020-01-17
[PAR - WDFT] Remy Pham - ongoing

kurosaek:
Everything works until I create an element.
As soon as I create an element, I can't delete items anymore and the quantity value returns to 0.

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -42,14 +42,37 @@
           <button class="btn btn-delete">Delete</button>
         </td>
       </tr>
+      <tr class="product">
+        <td class="name">
+          <span>IronShirt</span>
+        </td>
+    
+        <td class="pu">
+          $<span>15.00</span>
+        </td>
+    
+        <td class="qty">
+          <label>
+            <input type="number" value="0" min="0">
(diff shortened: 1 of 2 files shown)


Jepassaispar:
The reason is because you declare
```
var delButtons = document.querySelectorAll(".rm");
  delButtons.forEach(del => {
    let wrapper = del.parentNode;
    let parent = wrapper.parentNode;
    del.addEventListener("click", () => {
      parent.removeChild(wrapper);
      calcAll();
    });
  });
```
in the global scope, meaning your querySelector on the delButtons will be declared once, at the page load.
When you create a new Item, you are changing the innerHTML of the `cart` and `tbody` elements, so doing that the innerHTML of your element is changed and the querySelector won't be applied anymore on those elements (even if your are doing `+=` it will remove the selectors on the innerHTML already there).
A way of making it work would be to wrap everything inside a deleteButtons function that you will call once in the global scope and another time whenever you are creating a new item.
So everytime there is a created item, you will be able to select all the buttons and have actions on them.
The rest is good, well done :)

PR #1092, opened 2020-01-17
[PAR-WDFT-0120] - María.MUNIZAGA and Andy.TISBA ongoing

mmunizaga:


TatijanaRajcic:
Hey guys! Great work! 
There is only a small bug when i try to delete a product and then calculate the total again. If you look in the console, you see the following error:
Capture du 2020-01-22 14-22-23
This is due to the fact that when you delete an element, you are not removing it completely, just replacing its content by "".
A small change will fix that:

```
function deletingProducts(e){
  var totalAfter = document.querySelector('h2 span').textContent - e.currentTarget.parentElement.parentElement.querySelector('.subtot span').textContent

// HERE
  e.currentTarget.parentElement.parentElement.remove();
  document.querySelector('h2 span').innerHTML = totalAfter;

  var productsRemain = document.querySelectorAll('.product');
  var newTotal = 0;
  // productsRemain.forEach(p => updateSubtot(p));
  // productsRemain.forEach(p => newTotal += +p.querySelector('.pu span').textContent * p.querySelector('.qty input').value);
  // document.querySelector('h2 span').innerHTML = newTotal;
}
```

Nice work :)

PR #1093, opened 2020-01-17
[PAR-WDFT] - Kathleen D. ongoing

kathd:


TatijanaRajcic:
Hey Kathleen! Your code is perfect :) 

I don't have anything to add! You even thought of the calcAll inside the deleteProd function. It's perfect! :)))

PS: All of you added event listeners to all the delete buttons. There is another solution too: you could add an event listener to the parent element of the delete buttons. Here is an article about it if you want : 
https://davidwalsh.name/event-delegate

Great job!

PR #1161, opened 2020-01-28
done [PARWDFT - Andrew] ongoing

andrewnunezbibby:


TatijanaRajcic:
Hey Andrew! Thanks for your PR ! Nice work!

There is a slight confusion in your newItemBuilder function: you shouldn't target the class ".product" to retrieve the input values for the product we want to create, but the class "new". That's the class name of the line where you create a new product.

```
function newItemBuilder() {
  let product = event.target.closest(".new"); // CHANGE HERE
  console.log(product); 
  let name = product.querySelector("input[type=text]").value;
  let price = product.querySelector("input[type=number]").value;
  createItem(name, price)
}
```

- You need to add event listeners to the delete Buttons you create while creating new Products. To do that, just call your deleteItem function inside your createItem function:

`  newItem.querySelector(".btn-delete").onclick = deleteItem
`

Now, the delete buttons of the newly created lines work.

Closing the PR :) don't hesitate if you have questions!

PR #1177, opened 2020-02-16
[PAR - WDFT]- Shannon, done

tranquildoe:
Went to the review for this one.


clarabesset:
Same thing than the pizza lab, this isn't complete but I know you went to the review but if you don't understand something, let me know and we can do the lab again together.

PR #1233, opened 2020-03-06
LIS-PTWD-FEB-Catarina Laranjeiro

ctlaranjeiro:
@ironhack-LIS-PTWD
Having trouble on iter 4 and 5. 
5 is done except when I add a new product it goes on top of the list, though I'm using appendChild(). Only when I create a second one, the first one goes to the bottom, but the second remains on top of the list, and so on, and so on.

On iter 4, I'm having trouble targeting the remove buttons. It only keeps track of the button on top of the list.

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -33,7 +33,7 @@
           </td>
         </tr>
       </tbody>
-      <!-- <tfoot>
+      <tfoot>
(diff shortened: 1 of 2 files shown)


ironhack-LIS-PTWD:
Hey, 
Check your slack :) 
-Lissy

PR #1255, opened 2020-04-02
PAR WDFT Sam L ongoing

samlfair:


TatijanaRajcic:
Hey Sam!
Thanks for your lab! Looks nice! Since it's finished and it's working, I'll give you some tips on how to improve it:
- I think you overcomplicated things for your generateRowValues function :D It's nice because you practiced creating DOM elements from sratch, so this skill is useful! But it might get out of hands if you have to add a lot of inner HTML. To get things done quicker, you can: 1) create a DOM element with document.createElement as you did 2) and then modify its innerHTML as it was some text, but actually you add all the HTML you need in there. See this example:

```
function generateRowValues(e) {
  let productName = document.querySelector('.create-product input[type="text"]').value;
  let productPrice = document.querySelector('.create-product input[type="number"]').value;
  let newProduct = document.createElement('tr');
  newProduct.className = 'product';
  newProduct.innerHTML = 
  `
  <td class="name">
  <span>${productName}</span>
  </td>
  <td class="price">$<span>${productPrice}</span></td>
  <td class="quantity">
  </td>
  <td class="action">
  </td>
  `
  newProduct.querySelector('button.btn-remove').onclick = removeProduct;
  let tableBody = document.querySelector("tbody");
  tableBody.appendChild(newProduct);
  document.getElementById("calculate").onclick = calculateAll;
}
```


- To make your site more user friendly I would have executed "calculcateAll" after somebody deletes a row, so that the total gets updated automatically 
- If you would like to go further, you could make your site more user-friendly, for example by making your subtotals update every time somebody types something in the input (instead of waiting to hit the Calculcate Total button) :)

Otherwise it's really good :)

PR #1260, opened 2020-04-03
[PAR-FT-032020] Maxime Paolone

mxpln:
:frog:


ta-web-paris:
Nice job Maxime !!
We'll send a solution for the bonuses if you want to know how it's done :)

Nina

PR #1264, opened 2020-04-03
[PAR WDFT] Vladimir Leloup

Vlp90:
exercice done, without bonus


ta-web-paris:
Hey Vladimir !

Nice job !!

PR #1265, opened 2020-04-03
[PARIS-FT-042020] - Benjamin Dean

bdean05:
My cart is full


ta-web-paris:
Hey Ben, 
Nice job on the calculate functions ! 
We'll send a solution so you can understand the bonus iterations

Nina

PR #1266, opened 2020-04-05
PAR WDFT 202003 - Anne Patana - Week2 Day4

apatana:


ta-web-paris:
Hey Anne !

Nice job on the price calculation !
For the bonus section (create a new product) something is missing but we'll send a solution for everyone so you can read it and understand :)

Nina

PR #1267, opened 2020-04-06
PAR WDFT 0320 Teddy ROGERS

TeddyRod:


ta-web-paris:
Hey Teddy :)

Nice job on the subTotal ! We'll send a solution so you can have a look at the "calculate total price" and the bonus iterations.

Hit me up if you have questions !

Nina

PR #1391, opened 2020-06-21
REMOTE-WDPT-0602 [Sunitha Mannam]

SunithaMannam:
completed  all iterations and bonus


valmgisbert:
Very concise and organised code! Nice use of `Array` ! Good on showing different ways of doing the same thing, like with your eventListeners with your create-remove buttons. Remember to clean your code of console.logs. Excellent work!

PR #1392, opened 2020-06-21
[REMOTE-WDPT-0602] - Matthew Howard

mattalanhoward:


valmgisbert:
your ()s in `subtotal` aren't really needed since no other operation is within your variable, also remember to clean your console.logs in your code. Another way to create your new table row is to assign it's innerHTML that piece of HTML skeleton after you've created the element, still, very nice approach! Overall great job!

PR #1393, opened 2020-06-21
[REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:


valmgisbert:
you can refactor your iteration 3 even more; remember you can search selectors and their elements inside in a single DOM method. Wonderful job! Gold stars all around!

PR #1395, opened 2020-06-22
[Remote-PT-0620] Debora Crescenzo

NubivagantYansa:
All Iterations and Bonuses completed


valmgisbert:
You can assign the innerHTML straight after your querySelector without having to store it in a variable for a one-liner! Remember the spacing for a more organised and easier-to-read code. Excellent work!

PR #1396, opened 2020-06-22
[WDPT - JUNE/2020] - Willem Prins

willemprins93:
Finished.

Remove button doesn't work on newly created products, could not figure out why.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,22 +32,34 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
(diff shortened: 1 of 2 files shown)


valmgisbert:
In your iteration 4 you don't need to look for the values again, just point to the parent and the child; also, it can work without ()s while looking for your product child. Inside your `createProduct` you also gotta make sure that the removeBtn that you add into that new row inherits the same behaviour as any removeBtn; maybe that will solve your issue. Overall, amazing job!

PR #1397, opened 2020-06-22
[REMOTE-WDPT-0602] - ALEJANDRO GONZALEZ

alejanglez:
lab-dom-ironhack-cart

Iteration 2, not well explained in my opinion in the exercise.

Last iteration 5 was completly crazy, I had to collect info from a lot of places and students.

Not able to think something like that from now.

i think i will need a review of DOM methods, buttons .addenetlistener ...etc.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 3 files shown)


valmgisbert:
Remember to clean your console.logs,  nice use of the `.add()` ! Remember to organise and space your code accordingly to make it look cleaner. You did a great job! Just keep in mind that the btnRemove when you create them needs to inherit the same functionality as the others. You can also address your btnRemove and createLine inside a single window eventListener. Overall, you did very well! With some review it'll be even better!

PR #1400, opened 2020-06-23
[REMOTE-WDPT-0602] - Rodrigo Hernández

Rodrigohernandezsan:
I have had some difficulties.

Exercise 1, 2 and 3 moderate difficulty, but I have come to solve it.
Exercise 4 was solved thanks to stackoverflow, but I still don't have the logic of how to do it.
Exercise 5, I have not been able to do it.

Thank you very much

### DIFF
--- a/index.html
+++ b/index.html
@@ -50,7 +50,7 @@
           </td>
         </tr> 
       </tfoot>
-    </table>
+      </table> 
     <p class="calculate-total">
       <button id="calculate" class="btn btn-success">Calculate Prices</button>
     </p>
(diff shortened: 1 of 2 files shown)


valmgisbert:
Remember to clean your console.logs. There are ways to not have to call `Number()` on both times for your subtotal; look for what you want when you ask the values form the document. Try to look for alternate ways to do the 4th iteration as well; if you understand what's going on, and also based on your current logic, it's easier to find other ways to do it! Whenever you can, please work on finishing this.

PR #1401, opened 2020-06-23
[REMOTE-WDPT-0602] Cristina Matamoro

cris-developer:
Finalizadas iteraciones 1,2,3,4.
El Event Listener de la parte 5 parece no me crea la fila. No se porque.

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,12 +34,25 @@
         </tr>
         <!-- Iteration 2: Add more products here -->
       </tbody>
+      <tr class="product">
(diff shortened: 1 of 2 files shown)


valmgisbert:
Remember to clean your console.logs. You don't have eventListeners for create and delete buttons in your window. Be careful with starting your variables names capslocked, they might be recognised as something else and also not be referenced when you call them somewhere else. Good job so far! Whenever you can, come back to finish the final iteration.

PR #1402, opened 2020-06-23
[PT REMOTE WD 0620] Lynn W-B

lynn-w-b:
Done - so proud that I managed to complete it!

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,10 +32,22 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
-        <!-- Iteration 2: Add more products here -->
+        <tr class="product">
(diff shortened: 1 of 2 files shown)


valmgisbert:
If you're just using `i` in the loop, you don't need to create it outside. Work a bit on your spacing here, to have an even more organised code. Remember to clean up your console.logs. WOW LYNN, ALL THOSE ATTRIBUTES! INCREDIBLE! THE PATIENCE. THE DRIVE. I APPLAUD YOU (an alt way of doing this is directly adding an HTML skeleton of the row into an innerHTML method). This is amazing, I'm proud of you as well! Great job!

PR #1403, opened 2020-06-23
[REMOTE-WDPT-0602] Raquel Nascimento

RaquelNascimento96:
- First 3 iterations complete (still with doubts)
- Missing bonus


valmgisbert:
Good work so far! Be careful having code outside of the functions, because it might crash Jasmine. Whenever you can, come back and work on this.

PR #1441, opened 2020-07-14
WEB-REM-2007 Bodei Brouwer

BodeiBrouwer:
I feel like I am close but I'm still getting errors I cannot seem to figure out.

### DIFF
--- a/index.html
+++ b/index.html
@@ -23,16 +23,28 @@
           <td class="name">
             <span>Ironhack Rubber Duck</span>
           </td>
-          <td class="price">$<span>25.00</span></td>
+          <td class="price">€<span>25.00</span></td>
           <td class="quantity">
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
The map + reduce is actually a good approach. Issues you were having were:
- Line 19. In order for .map to work you need to actually convert the nodeList (products) into a regular array. You can do this with spread operator, but this is kinda advance stuff we will learn in module 2. in the end it will look like this: 
let subTotals = [...products].map((product) => {
    return updateSubtotal(product);
 });
- Line 23. You want to save the .reduce result inside the variable total. so: let total = subTotal.reduce ...
- Line 29. There is a typo, it is total-value not total-values.
Still, good job and clever logic using those advanced methods! Keep it up.

PR #1442, opened 2020-07-14
Web Development - Remote- july2020 Joel Roca Martinez / shopping-chart-LAB

JoelRocaMartinez:
I´ve cried blood to complete the 3 Iterations

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,6 +32,19 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
+          <td class="name">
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Tears of blood was the answer we were looking for :)
Code looks fine and I can see the great effort in all of those commented console.log, however this code will only run with two products. You want to make it dynamic, and with a loop, allow it to run even if multiple products exist. Check the solutions and see the logic of working with functions and loops to allow a more modular and dynamic code. If you have time, and after checking the solution, try to go over it again.

PR #1443, opened 2020-07-14
[WDFT-REM-2007] Maëva Merrou

maevamerrou:
Hey, 
Didn't finish as my code doesn't work and I am quite confused about this new topic.
Hopefully, I will get to understand better in the coming days.
Maeva

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,6 +32,32 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
+          <td class="name">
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
It looks good! I guess you only stumble a little in iteration 3, in line 25, product doesn't exist in this function, only in the updateSubtotal function. So what you want to do is return the subtotal in the updateSubtotal function and then use it when you call it in the calculateAll function.

PR #1444, opened 2020-07-14
[WEB-REM-2007] ALVARO SANCHEZ LAMADRID

AlvaroSLam:
All mandatory iterations done


IH-WebDev-TA-Remote:
Good job! it works, although part of mandatory was also doing the total amount. But still, good work. Keep it up!

PR #1445, opened 2020-07-14
[WEB-REM-2007] Marta Lourido

MartaLourido:
I thought I'd never say it, but it's done!

Pretty challenging, but very fun to learn how things work.

I haven't managed to make the bonus.


IH-WebDev-TA-Remote:
Very good job! all seems to be working correctly :)
Just one thing I note, in the html in line 28 you added an id in front of the type, and that is breaking the correct input format.

PR #1446, opened 2020-07-14
[WEB-REM-2007] RAFAEL DE LAZARO

rlaz133:
When appending the new product to the list, the class .product is removed from it, but its children kept their classes. I still don't completely understand this behavior.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,22 +32,48 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
+          <td class="name">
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good work on the lab! all iterations done.
Whenever you create the element you need to assign the class via the setAtribute method. I see you were trying to assign the class via innerHTML but remember the atribute is not inside the html, it is part of the tag. Now the inner elements you are setting the classes with simple template literals. But it doesn't work on the main element you create.

PR #1447, opened 2020-07-15
[WDFT-REM-2007]Gonzalo

gonzalo-development:


IH-WebDev-TA-Remote:
There is a syntax error in lines 61 and 62 that are breaking everything. Remember that to use string interpolation you need to use backticks(´) and not regular quotes (').
After fixing that everything runs well. And one bonus done. Good job! Keep it up.

PR #1449, opened 2020-07-22
Marcella - Amsterdam wd pt June2020

allecramarcella:
In iteration 5 I wasn't able to have the new product that is added to the cart pushed to the bottom of the cart. It now adds  at the top. How can I do this?

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -7,6 +7,7 @@ function updateSubtotal(product) {
   const quantityValue = quantity.valueAsNumber;
   
   price = price.innerHTML;
+  // console.log(price)
 
   const subTotal = price * quantityValue;
   // console.log(subTotal)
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -2,41 +2,119 @@
...
+  
+
+  const productLine = document.querySelector('.product');
+  const parentProductLines = productLine.parentNode
+  console.log(productLine)
+  console.log(parentProductLines)
+  const cloneProductLine = productLine.cloneNode(true);
+
+  
+  parentProductLines.appendChild(cloneProductLine); //How can I make the new line appear as the last product??
```
guido-an, line 93: The problem is that you are correctly appending the <tr> as last child but then your are changing the wrong innerHTML. After you append the cloneProductLine you are changing the name and the price of the first product. 

You should first create the new element and then append it:
`function createProduct () {`
  `const nameNewProduct = document.getElementById('name-new-product').value`
  `const priceNewProduct = document.getElementById('price-new-product').valueAsNumber`

  `const productLine = document.querySelector('.product')`
  `const parentProductLines = document.getElementById('productsRows')`
 ` const cloneProductLine = productLine.cloneNode(true)`
  `cloneProductLine.querySelector('.name span').innerHTML = nameNewProduct`
 ` cloneProductLine.querySelector('.price span').innerHTML = priceNewProduct`

  `parentProductLines.appendChild(cloneProductLine)`
`}`

PR #1452, opened 2020-07-22
[WDPTAMS062020] Anna Dorenskaia

AnnaDore:
Unfortunately I coundn't win a fight with the bonus iteration 4 and would be happy to know a solution for it if it is possible


guido-an:
Really good Anna!

PR #1487, opened 2020-08-19
WDTF-Berlin- Yoseob Shin - done without last bonus

shinyo627:


shinyo627:
finished whole thing including bonuses

PR #1521, opened 2020-09-03
[PAR-WDFT-248] Fabien Marchand

fabienMarchand:


Fanny-Sbn:
Hello Fabien, thanks for your lab. Just a few comments. 

-In your function calculateAll() no need to use a spread operator and you can directly use a forEach on productList.
function calculateAll() {
  let productList = document.querySelectorAll(".product");
  let totalProduct = 0;
  productList.forEach( product => totalProduct += updateSubtotal(product));
}
-For your removeProduct() function, target.parentNode.parentNode.parentNode is not really elegant try to look for closest 😉
-Line 63 no need of a spread operator
-You forget to update the total price when removing a product 😉

Good job 💯

PR #1523, opened 2020-09-03
[PAR WDTF] Laurent Vuillaume

figura01:
iteration 1, iteration 2, iteration 3 done

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,41 +1,57 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width" />
-    <link rel="stylesheet" href="./css/style.css" />
-    <title>Ironhack Cart</title>
-  </head>
-  <body>
-    <h1>Ironhack Cart</h1>
-    <table id="cart">
-      <thead>
-        <tr>
-          <th>Product Name</th>
-          <th>Unit Price</th>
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hey Laurent, thanks for submitting your lab. 
In your function updateSubtotal(), you initialize your variable price, quantity, and unitPrice and then reassign them a value straight after. Why not doing both at same time? 
You don’t need to use toFixed() and convertValue() here. Like this you’re saving a few lines:

  let quantity = product.querySelector('.quantity input').value;
  let unitPrice = product.querySelector('.price span').innerHTML;
  let subtotal =  product.querySelector('.subtotal span');
  let price = (quantity * unitPrice);
  subtotal.innerHTML = price;
  return price;
}

For your function calculateAll(), it’s working but instead of a for loop you could have used a .forEach :
products.forEach(product=> total += updateSubtotal(product));
Again, no need to use toFixed() here. 
-For your removeProduct() function, target.parentNode.parentNode.parentNode is not really elegant try to look for closest 😉
Instead of writing  targetParentNode.lastChild.lastChild.childNodes[1].addEventListener("click", removeProduct) , create a new variable that targets the remove button and add an event listener on it : 
const removeProductBtn = tplTr.querySelector(".btn-remove");
removeProductBtn.onclick = (evt) => removeProduct(evt);

Good job Laurent :100

PR #1530, opened 2020-09-03
[PAR-WDFT-248] Bouziane Bey

Bouziane95:


Fanny-Sbn:
Hello Bouziane, thanks for submitting your lab. 

**Iteration 1** 
-
👍 

**Iteration 2**
-
In your calculateAll() function you have another function named totalReducer() that is never called, so you can remove it...
Your solution is working but you could have also declared a variable total before using a forEach method and add directly the updateSubtotal(product) to it.

function calculateAll() {

const allProducts = document.querySelectorAll(".product");
let displayTotal = document.querySelector('#total-value span');

let total = 0;
allProducts.forEach(product => totalPrice += updateSubtotal(product));
displayTotal.textContent = total;
}

For your removeProduct() function, parentELement.parentElement.outerHTML is not really elegant try to look for closest 😉

Have you tried iteration 5 ? 😉

PR #1531, opened 2020-09-03
[PAR WDFT 248] Jorell Langmar

JorellLangmar:
- I'll change the create button style when I'll have time.
- And I'll try to add some EventTarget on the quantity to launch CalculateAll directly etc.

### DIFF
--- a/index.html
+++ b/index.html
@@ -40,6 +40,19 @@
           </td>
           <td class="price">$<span>45.50</span></td>
           <td class="quantity">
+            <input class="input-qtty" type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
+        <tr id="hidden-product" style="display: none;">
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Jorell, thanks for submitting your lab. 

**Iteration 1**
-
No need to declare subtotal.innerText as an empty string and then as the subtotal, you can directly attach the subtotal to the variable:

subTotal.innerText =subTotalPrice;

**Iteration 2 & 3**
-
It's working but you could have been more concise and  declared a variable totalValue before your forEach method and add directly the updateSubtotal(product) to it.

function calculateAll() {

const allProducts = document.querySelectorAll('.product');
let totalValue = 0;
allProducts.forEach(productUnit => totalValue += updateSubtotal(productUnit));

const totalPrice = document.querySelector('#total-value span');
totalPrice.textContent = totalValue;
}

**Iteration 4**
-
For your function removeProduct, target.parentElement.parentElement.parentElement is not really elegant, have a look to closest() method. 😉

**Iteration 5**
-
Your remove buttons for the created items don't work, you just need to change one thing 😉 

Good job overall 💯

PR #1532, opened 2020-09-03
[PAR-WDFT-248] Nicolas Enault

nenault:


Fanny-Sbn:
Hello Nicolas, thanks for submitting your lab. 

**Iteration 1**
-
👍 

**Iteration 2 & 3**
-
 👍 You don't need line 21 to 23, you're doing twice the same thing with the second forEach(). 

**Iteration 4**
-
👍 Nice use of the while loop ! You could have used closest() also 😉 

**Iteration 5**
-
👍 

Super good job 💯

PR #1534, opened 2020-09-03
[PAR-WDFT-248] Laura Caillaux

Lcdevi:


Fanny-Sbn:
Hey Laura, thanks for submitting your lab. 

**Iteration 1**
-
👍 

**Iteration 2&3**
-
It's working but you could have done the same in a much shorter way. No need to create variables you won't use outside of this scope in that case. 

You could directly go for : 

function calculateAll() {

const allProducts = document.getByClassName('product');
let totalValue = 0;
[...allProducts].forEach(element => total+= updateSubtotal(element));

document.querySelector('#total-value span').innerHTML = total;
}

**Iteration 4**
-
For your function removeProduct, target.parentNode.parentNode is not really elegant, have a look to closest() method. 😉

**Iteration 5**
-
👍 
As you use a querySelectorAll for your variable allRemoveButton, you don't need to transform in an array allRemoveButton, you can directly use forEach on it 😉 

Good job 💯

PR #1548, opened 2020-09-22
[REM 2009] Ross Homaidan

ibirossi:
I was stuck on Iteration 3, without help I would never have got the answer.


IH-WebDev-TA-Remote:
Everything working up to iteration 3. Good job!

Don't worry, the important thing is that you understand everything that is happening in your code up to iteration 3. The rest comes with practice and today you will be practicing even more DOM with the lab.

PR #1550, opened 2020-09-22
[REM 2009]Marta Obregon

MartaObregon:


IH-WebDev-TA-Remote:
Alright! everything working up to iteration 3!

I also see you were on a good start for iteration 4, there were just some wrong variable names being thrown that were breaking the code but overall good approach. This is what you needed there, try to compare it and see what went wrong in your code. Of course this will only run the function removeProduct, then you needed to do the code for actually deleting the element.

```
const allRemoveButtons = document.querySelectorAll(".btn-remove")
  allRemoveButtons.forEach((btn)=>{
      btn.addEventListener('click', removeProduct)
  })
```

Still overall good job, keep it up!

PR #1552, opened 2020-09-22
[REM - 2009] Vykintas Petrikas

VykiPetr:
COMPLETED! I have a question about a certain part in the createProduct, it is too long to write out! I hope ill remember to ask in class!!

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,22 +32,48 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
+          <td class="name">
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
The create part is supposed to be long, the only suggestion to decrease it a little bit is to assign the addEventListener only to the element created, not to all the elements again. Same as you use querySelector and querySelectorAll to the document (all of DOM) you can use them with a single DOM element and it will look only through that one.

Overall good job! everything working!

PR #1553, opened 2020-09-22
[REM-2009] Sandra Mendez

scmendez:
**question on iteration 5**

this is what i have - 
    let productNameInput = document.querySelector('.create-product input[type="text"]');
    let unitPriceInput = document.querySelector('.create-product input[type="number"]');
...
      <td class="name"><span>${productNameInput.value}</span></td>
      <td class="price">$<span>${unitPriceInput.value}</span></td>
...
    productNameInput.value = ''
    unitPriceInput.value = ''

but I tried this, it didn't work, and I don't understand why :(
    let productNameInput = document.querySelector('.create-product input[type="text"]').value;
    let unitPriceInput = document.querySelector('.create-product input[type="number"]').value;
...
      <td class="name"><span>${productNameInput}</span></td>
      <td class="price">$<span>${unitPriceInput}</span></td>
...
    productNameInput = ''
    unitPriceInput = ''

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,59 +1,75 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width" />
-    <link rel="stylesheet" href="./css/style.css" />
-    <title>Ironhack Cart</title>
-  </head>
-  <body>
-    <h1>Ironhack Cart</h1>
-    <table id="cart">
-      <thead>
-        <tr>
-          <th>Product Name</th>
-          <th>Unit Price</th>
-          <th>Quantity</th>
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
That is odd, it is actually working for me. the problem I am seeing is that the html format you are using when creating the DOM element is not the same as the original product html. Try this with your same code and it should work:

````
newProduct.innerHTML = `
    <td class="name">
      <span>${productNameInput.value}</span>
    </td>
    <td class="price">$<span>${unitPriceInput.value}</span></td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
    `;
````

Overall though, very good job! everything else is working good :)

PR #1554, opened 2020-09-22
[REM-2009] Sumiya Ushiro

sumi0820:
I managed to complete all iteration including bonus. However, i was stuck on remove button for newly added item. 

The problem I faced was removeProduct function does not work to new items as the new items do not have the parent node, which is <tbody>, so I was unable to use removeChild(). 

The approach I took is to create new event listener inside the create item function. Then I managed to access <tbody> of the new item and remove it. I feel there would be the way to re-use removeProduct function, but I do not figure it out. Kindly advise if there is any way to do that.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "git.ignoreLimitWarning": true
+}
--- a/index.html
+++ b/index.html
@@ -19,6 +19,7 @@
         </tr>
       </thead>
       <tbody>
+
         <tr class="product">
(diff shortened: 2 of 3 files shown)


IH-WebDev-TA-Remote:
Probably you had some syntax error before, you do can simply re-use the removeProduct function, just replace lines 96 and 97 with the line below and it should work!

`removeProduct(event, removeBtn)`

Still the approach you did was good and the lab overall really well done! keep it up! :)

PR #1555, opened 2020-09-22
[REM 2009] Aleksandra Tanasijenko

alekstana:
got stuck at bonus iteration 4 :/

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,6 +32,19 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Beach Towel</span>
+          </td>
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good, all iterations up to the third working good! I also see you tried hard on iteration 4 and even had a good approach! The problem was you added the addEventListeners inside the function removeProduct, however this function is never being called anywhere in the code.

The correct way to go around this was to do the addEventListeners inside the last addEventListener "load" and have those run the removeProduct function for each button, then inside the removeProduct function you just run the two lines to get the parentNode and remove the child. If this is complex to understand check the solution we will send later.

Still overall good work and goo effort on the bonus, keep it up!

PR #1556, opened 2020-09-23
[REM-2009] Santiago Andrés Ardiles Páez

SantAndresP:
I feel that lines **_65_** and **_66_** are kind of redundant, but I couldn't think of another solution.

**_65_** `const remove_product_button = new_product.querySelector(".btn-remove");`
**_66_** `remove_product_button.addEventListener("click", removeProduct);`

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,8 +6,10 @@
     <link rel="stylesheet" href="./css/style.css" />
     <title>Ironhack Cart</title>
   </head>
+
   <body>
     <h1>Ironhack Cart</h1>
+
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
those lines might feel redundant but is the best approach, you are creating a completely new element in DOM so it makes sense that if you want them to do something you assign the addEventListener.

Overall amazing job, I see you even managed to clear the total when deleting elements with a subtotal calculated! Keep it up!

PR #1557, opened 2020-09-23
[REM-2009] Mohammed Imran

14imran:
completed three iterations.


IH-WebDev-TA-Remote:
Very nice solution, and very clever to use Array.from, although you would mostly use that in order to be able to use other array methods on it. forEach works on nodelists exactly the same as for arrays.

Overall, good job, just don't be afraid to try those bonuses!, keep it up!

PR #1558, opened 2020-09-23
[ REM 2009 ] Jordane Ruiz

jordaneruiz:
Need to seriously work again on DOM


IH-WebDev-TA-Remote:
That is good! you will have time to continue working with DOM with today's lab :) 

Still, I see you finished the three iterations! the most important thing is that everything that is happening there is making sense.

So goo job! Keep it up!

PR #1559, opened 2020-09-23
[REM-2009] Juan Jose

JJCExcelsior:


IH-WebDev-TA-Remote:
The code logic looks good but there are some errors happening that are breaking the code, also the approach inside the removeProduct function is good, but the problem is that the function is never been called so it is never running. Try to go back to the lab and fix those issues to make it work, so at least the first 3 iterations work.

PR #1607, opened 2020-10-20
[FTAMS102020] Nina van den Braak

ninsajala:
Finished the iterations, need to do a double check to make things cleaner

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -27,6 +27,7 @@ function calculateAll() {
 function removeProduct(event) {
   const target = event.currentTarget;
   itemList.removeChild(target.parentNode.parentNode)
+  calculateAll()
 }
 
 function activateRemoveButtons() {
--- a/js/index.js
+++ b/js/index.js
@@ -1,23 +1,23 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector('.quantity input')
+  const subTotal = product.querySelector('.subtotal span')
+  let totalValue = parseInt(price.innerHTML) * parseInt(quantity.value)
+  subTotal.innerHTML = totalValue
+  return totalValue
 }
 
 function calculateAll() {
-  // code in the following two lines is added just for testing purposes.
-  // it runs when only iteration 1 is completed. at later point, it can be removed.
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
-  // end of test
+  let allProducts = document.querySelectorAll('#cart tr.product')
+  let endTotal = 0
 
-  // ITERATION 2
-  //... your code goes here
-
-  // ITERATION 3
-  //... your code goes here
+  for (let i = 0; i < allProducts.length; i++) {
+    endTotal += updateSubtotal(allProducts[i])
+  }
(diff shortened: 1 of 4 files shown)


js/index.js
```diff
@@ -1,42 +1,75 @@
-// ITERATION 1
+//Global HTML selectors
+const displayTotal = document.querySelector('#total-value span')
```
wax-bear, line 2: "displayTotal" is only used in the function "calculateAll". Thus, I would only declare it inside that function. I would also suggest thinking about a more descriptive name of what "displayTotal" actually is. Perhaps "totalValueSpanEl" or something like that.

wax-bear, line 2: "displayTotal" is only used in the function "calculateAll". Thus, I would only declare it inside that function. I would also suggest thinking about a more descriptive name of what "displayTotal" actually is. Perhaps "totalValueSpanEl" or something like that.

js/index.js
```diff
@@ -1,42 +1,75 @@
...
+const itemList = document.querySelector('tbody')
+const newProductName = document.querySelector('.product-name input')
```
wax-bear, line 4: As far as I can see, these variables are only used in one function. Thus, I would declare them over there.

js/index.js
```diff
@@ -1,42 +1,75 @@
...
+const newProductPrice = document.querySelector('.product-price input')
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector('.quantity input')
+  const subTotal = product.querySelector('.subtotal span')
+  let totalValue = Number(price.innerHTML) * parseFloat(quantity.value)
```
wax-bear, line 11: Is totalValue ever reassigned?

js/index.js
```diff
@@ -1,42 +1,75 @@
...
+const itemList = document.querySelector('tbody')
+const newProductName = document.querySelector('.product-name input')
+const newProductPrice = document.querySelector('.product-price input')
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector('.quantity input')
```
wax-bear, line 9: Please be consistent in the use of semicolons.

js/index.js
```diff
@@ -1,42 +1,75 @@
...
+            <span>${newProductName.value}</span>
+          </td>
+          <td class="price">$<span>${newProductPrice.value}</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>`
```
wax-bear, line 58: I would insert a blank line after this one.

js/index.js
```diff
@@ -1,42 +1,75 @@
...
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
 
-  //... your code goes here
-});
+  activateRemoveButtons()
+
+  const createButton = document.querySelector('#create')
+  createButton.addEventListener('click', createProduct)
+});
```
wax-bear, line 75: Great stuff Nina!
The UI works as expected.
I placed some comments regarding naming, style and const vs. let.

PR #1617, opened 2020-10-22
[FT-AMS-102020] Emad Ghnim

Emad-doz:


js/index.js
```diff
@@ -1,42 +1,90 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
+  //console.log('Calculating subtotal, yey!');
```
LloydC, line 4: I would recommend removing any code/comment which no longer serves a purpose, it'll keep your code minimal and easier to read through

js/index.js
```diff
@@ -1,42 +1,90 @@
...
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
+  //const singleProduct = document.querySelector('.product');
+  //updateSubtotal(singleProduct);
   // end of test
 
   // ITERATION 2
-  //... your code goes here
+  const productsArray= document.getElementsByClassName('product');
+  console.log(productsArray);
```
LloydC, line 22: Again, once you have confirmed it works you can remove your `console.log()`

js/index.js
```diff
@@ -1,42 +1,90 @@
...
+
+
+  // ... your code goes here
 }
 
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
 
+  const removebutton = document.getElementsByClassName('btn-remove');
```
LloydC, line 82: It looks that you are following "camel-case" style notation, so your variable should be named `const removeButton`

js/index.js
```diff
@@ -1,42 +1,90 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
+  //console.log('Calculating subtotal, yey!');
+  const price = (product.querySelector('.price span').innerHTML);
```
LloydC, line 5: You don't need extra parentheses, you could just write it `const price = product.querySelector('.price span').innerHTML`

PR #1621, opened 2020-10-22
[FT-AMS-102020] Yann Quintin

YannQuintin:
This is the working solution for the cart.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,42 +1,92 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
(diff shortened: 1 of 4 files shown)


js/index.js
```diff
@@ -1,42 +1,92 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
+  console.log("Calculating subtotal, yey!")
```
LloydC, line 4: After you're done and submit, I recommend removing any code/comments which doesn't serve a purpose. It'll make your code minimal and easier to read

PR #1622, opened 2020-10-22
[FTAMS102020] Irene Asensio

ihatepineapple:
Missing bonus iteration 5

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,23 +1,43 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
+//Get price and quantity values and convert them to numbers
 
-  //... your code goes here
+  const price = product.querySelector('.price span').innerHTML;
+  const quantity = product.querySelector('.quantity input').value;
+  console.log(Number(price), Number(quantity));
+
+// Get the subtotal value
+// Render the subtotal value in element class ".subtotal"
+
+  const subtotal = Number(price) * Number(quantity);
+
+  const subtotalElement = product.querySelector('.subtotal span');
+  subtotalElement.innerHTML = subtotal;
+  return subtotal;
+
 }
 
 function calculateAll() {
-  // code in the following two lines is added just for testing purposes.
-  // it runs when only iteration 1 is completed. at later point, it can be removed.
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -1,42 +1,87 @@
...
+
+  for (cartItemElement of cartItemElements) {
+    totalValue += updateSubtotal(cartItemElement);
+  }
+  
+  totalValue;
   // ITERATION 3
-  //... your code goes here
+
+  const totalValueSpan = document.querySelector('#total-value span').innerHTML = totalValue;
```
LloydC, line 38: Line 35 is unnecessary (totalValue has been defined and modified via line 29-33)

Since you're not looking to use the stored value then you don't need to store it in a variable. You can just write  `document.querySelector('#total-value span').innerHTML = totalValue;`

js/index.js
```diff
@@ -1,42 +1,87 @@
...
-  //... your code goes here
+  
+  // const createProductRow = document.getElementsByClassName('create-product')
+
+  const createProductRowFoot = document.querySelector('tfoot')
+  const newProductNameInput = createProductRowFoot.querySelector('input').value
+
+  crea
+
+
```
LloydC, line 72: It looks like you were on the right track, here's a complete working solution
```
function createProduct() {
  const createProductRow = document.querySelector(".create-product");
  const newProductNameInputValue = createProductRow.querySelector("input").value;

  const newProductPriceValue = createProductRow.querySelector("input[type='number']");
  const newProductPrice = Number(newProductPriceValue.value).toFixed(2);

  const newProductRow = document.createElement("tr") 
  newProductRow.className= "product";
   newProductRow.innerHTML = `<tr class="product">
    <td class="name"> <span>${newProductNameInputValue}</span></td>
    <td class="quantity"> <input type="number" value="0" min="0" placeholder="Quantity" /></td>
    <td class="action"><button class="btn btn-remove">Remove</button></td>
   </tr>  `;

const productTableBody = document.querySelector("tbody");
productTableBody.appendChild(newProductRow);

const newProductRemoveButton = newProductRow.querySelector(".btn-remove");
newProductRemoveButton.addEventListener("click", removeProduct);
}

PR #1626, opened 2020-10-23
[FTAMS102020] Sophie Tichelaar

Sophietic:
done

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,33 +1,49 @@
 // ITERATION 1
 
-function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
 
-  //... your code goes here
+function updateSubtotal(product) {
+  
+  const price = product.querySelector('.price span').innerHTML 
+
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -1,33 +1,49 @@
...
+
+  const productTable = productRow.parentNode;
+
+  productTable.removeChild(productRow);
+
+  calculateAll();
+}
 // ITERATION 5
 
 function createProduct() {
```
LloydC, line 49: Good work, I don't see anything worth mentioning with your code(maybe use less whitespace )

I will add a working solution for iteration 5, let me know if you have any questions
```
function createProduct(event) {

  newTableRow.innerHTML = `
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  `;

  // clean the fields
  newProdNameInput.value = '';
  newProdPriceInput.value = 0;
}

PR #1627, opened 2020-10-23
[FTAMS102020] AlejandroPliakasBenitez

apliakas:


js/index.js
```diff
@@ -1,23 +1,32 @@
...
+  for (item of availableItems) {
+    total += updateSubtotal(item);
+  }
 
   // ITERATION 3
-  //... your code goes here
+  document.querySelector('#total-value span').innerHTML = total;
 }
 
 // ITERATION 4
```
LloydC, line 32: Good work, I don't see any issues with your code/solution

I'll add a working solution for iteration 4 and 5 so you can check the logic for it in your own time
```
// ITERATION 4
  // ...
  const row = target.parentNode.parentNode;
  //   console.log('row: ', row);
  const parent = row.parentNode;
  //   console.log('parent: ', parent);
  parent.removeChild(row);

  calculateAll();
}

// ITERATION 5

function createProduct() {

  newTableRow.innerHTML = `
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  `;

  // clean the fields
  newProdNameInput.value = '';
  newProdPriceInput.value = 0;
}

PR #1683, opened 2020-10-27
[WDFTRM 20201027] Rosana Batista

rosanabatista:
iteration 5 to go

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -5,7 +5,7 @@ function updateSubtotal(product) {
   const price = product.querySelector('.price span').innerText;
   const quantity = product.querySelector('.quantity input').value;
   const subtotal = price * quantity;
-  product.querySelector('.subtotal').innerText = subtotal;
+  product.querySelector('.subtotal').innerText = `$${subtotal}`;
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -2,41 +2,56 @@
 
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
+  const price = product.querySelector('.price span').innerText;
+  const quantity = product.querySelector('.quantity input').value;
+  const subtotal = price * quantity;
+  product.querySelector('.subtotal').innerText = subtotal;
```
filipe-freire, line 8: In this case, you could target the span elements inside the subtotal class, in order to leave the dollar sign next to the subtotal value! Ex: $ <td class="subtotal"> $<span>40</span></td>. This way, you'd only change the value inside the span 😀

filipe-freire (commented):
Very nice work Rosana! One more lab in the bag! 💪

js/index.js
rosanabatista: ok, just changed, thanks!

PR #1752, opened 2020-11-19
[PAR-WDFT-911] Marine SANJUAN

marineks:
Hello,

Please find my submission for today's lab.

My biggest points of struggle were: 
1) realizing that I queried the element in the document and not the product in Iteration 2
2) in Iteration 3, my subtotals were not added but put side by side. It took me time to realize that in Iteration 1 the subtotal was a number in a string (so a string...)

Elements I understood better:
- querySelector, parents elements and children
- how to use innerHTML for data types other than a string
- how to debug better with console.log
- that I could copy an HTML collection to convert it into an array with [...] and use HOF on it

Best,
Marine

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
         </tr>
       </thead>
       <tbody>
-        <tr class="product">
+        <tr class="product" >
           <td class="name">
             <span>Ironhack Rubber Duck</span>
           </td>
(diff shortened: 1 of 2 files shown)


VProum:
Hello Marine !

Yeah, when dealing with javascript, always keep in mind what exactly is the data you are working on ;)
never try to guess, always console.log it to be sure, and keep in mind the type also.

### Iteration 1: updateSubtotal & Iteration 2: calculateAll()
Clean job 👍
your code is quite difficult to read because of all the comment in it, you can remove the console.log once you are sure it is working as planned :D

### Iteration 3: Total
Good Job 👍 
maybe you can try with a reduce? ;)

I'm sure you can handle the bonus ! :D
keep up the good work!

PR #1761, opened 2020-11-20
PAR-WDFT-112020 Kenton Cargile

KentonCargile:


Fanny-Sbn:
Hello Kenton, thanks for submitting your lab.

On my side nothing is working. The first thing I did was to inspect in my browser and it showed an error "Unexepected end of input".  How did you check that your code was working? You need to check that your JS is valid first of all.

PR #1764, opened 2020-11-20
[PAR-WDFT-911] - Thomas LAMY

thomasl4012:
Lab done with all the bonuses.

There is a bug the created rows can't be removed. 
It's coming from the fact that the click handler on theses buttons is set on load so the new created buttons don't get the event handler.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,22 +32,41 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+
         <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Beach Towel</span>
+          </td>
+          <td class="price">$<span>12.50</span></td>
+          <td class="quantity">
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Thomas, thanks for submitting your lab.

Here are a few comments :

Iteration 1: Good 👍

Iteration 2 &3: It's working but it could be simplified, one HOF is enough to do everything. 

function calculateAll(){
let allProducts = document.querySelectorAll(".product");
let total = 0;
allProducts.forEach(product => total = updateSubtotal(product));
const totalPrice = document.querySelector("#total-value span");
totalPrice.innerHTML = total;
}

Iteration 4: Better than using target.parentNode.parentNode try to have a look for closest ;)

Iteration 5: Good, to handle the remove button, you just need to add an add evenlistener on the remove button when you create a new row -> cells.querySelector(".btn-remove").onclick = removeProduct;

Good job overall Thomas 💯

PR #1765, opened 2020-11-20
[PAR-WDFT-911]YUMIAO

yuuumiao:


VProum:
Hello Yu!

Excellent work! the bonus are working quite well and your code is clean
i see that you console.log your value, this is a good habit :)

### Iteration 1: updateSubtotal & Iteration 2: calculateAll()
Clean job 👍 you even put the extra work of converting the const into number!

### Iteration 3: Total
Good Job 👍 nice use of reduce!

### Iteration 4: Removing a product
Good job 👍 nice use of closest :D
but maybe you want to update the total price
(call the  calculateAll()  function)

### Iteration 5: Creating new products
there is a lot to say here :
- your remove button on the created product is not working, you have to add an eventlisterner to it
- update the total price at the end
- when you add your product, don't forget to set the quantity and recalculate the subtotal of this product, right now the quantity of the new product is the same of the 1rt product and it is a bit strange

PR #1766, opened 2020-11-21
[PAR -WDFT-911] Anojan Mylvaganam

anojanmyl:
didn't do the iteration 5 and the iteration 4 i don't know if there's a better way since 'event' is deprecated

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,10 +32,22 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
-        <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Beach Towel</span>
+          </td>
+          <td class="price">$<span>12.50</span></td>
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Anojan, thanks for submitting your lab.

Great job! Here are a few comments :

Iteration 1: Good 👍 but just try to change your innerHTML of your tag on one line and return only the total. 
totalPrice.innerHTML = total;
return total;

Iteration 2 &3: Good 👍 You can write your HOF on one line 

function calculateAll() {
  let allTotal = document.querySelector('#total-value span');
  let sum = 0;
  let allProducts = [...document.getElementsByClassName('product')];
  allProducts.forEach(product => sum += updateSubtotal(product));
  allTotal.innerHTML = sum;
}

Iteration 4: Event is not deprecated but it's better to use closest instead of parentNode.parentNode.parentNode. Try to have a look ;)

PR #1768, opened 2020-11-22
[PAR WDFT 911] Calypso Asline

Calyaln:
Couldn't find a way to make bonus 2 work

### DIFF
--- a/index.html
+++ b/index.html
@@ -27,15 +27,27 @@
           <td class="quantity">
             <input type="number" value="0" min="0" placeholder="Quantity" />
           </td>
-          <td class="subtotal">$<span>0</span></td>
+          <td class="subtotal">$<span>0</span></td>          
           <td class="action">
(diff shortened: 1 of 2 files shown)


VProum:
Hello Calypso!

All's good 👍
nice usage of HOF :D

### Iteration 1: updateSubtotal
Clean job 👍

### Iteration 2: calculateAll()
Good Job 👍 

### Iteration 3: Total
Good job 👍

### Iteration 4: Removing a product
Good job 👍 you think of updating the price after the delete!

### Iteration 5: Creating new products
step to do the iteration 5 :
- select the two inputs with a query selector to obtain their values
- Create a HTML tr object with the correct structure (or clone one)
- Change the innerHTML of the new tr object with the value of the query selector
- Add an event listener on the delete button of the tr
- Add the tr to your HTML
- Update the price

PR #1769, opened 2020-11-22
[PAR-WDFT-911]Gwervorc'h Mahéo

Billybillycode:
need to comeback on it after the video game challenge

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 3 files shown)


N10T:
Hello Gwerborc'h,

Iteration 1
-

Good job 👍 

Iteration 2
-

You targeting the wrong class, you need to find the "product" class to find the good elements then update it
```
  const allProducts = [...document.getElementsByClassName("product")];
  allProducts.forEach((prod) => {
    updateSubtotal(prod);
  });
```

Iteration 3
-

Good job 👍  and you used a reducer... in this case you don't need to map before
```
  const sumSubTotal = [...document.querySelectorAll(".product .subtotal span")]
    .reduce((acc, curValue) => acc += Number(curValue.textContent), 0);
```

Yes, you can try to do the bonuses of this lab ;)

PR #1770, opened 2020-11-22
PAR-WDFT 911- Etouma Bela

Etouma-bela:
I failed to calculate the total price and the remove button does't work. Overall I have issues with loopings

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,6 +33,19 @@
           </td>
         </tr>
         <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Beach Towel</span>
+          </td>
+          <td class="price">$<span>12.50</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
       </tbody>
       <tfoot>
         <!-- <tr class="create-product">
--- a/js/index.js
+++ b/js/index.js
@@ -1,31 +1,45 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
+  console.log("Calculating subtotal, yey!");
 
-  //... your code goes here
+  const price = product.querySelector(".price span");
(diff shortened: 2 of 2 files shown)


Fanny-Sbn:
Hello Etouma, thanks for submitting your lab. 

Here are a few comments :

Iteration 1: Good 

Iteration 2: Good 

Iteration 3 : Have you opened your console? It throws two errors : i is not defined line 32, that means you're trying to use a variable that is not defined at this point and cannot set property innerHtml of undefined line 32 which means your totalPrice is not what you think, you made a mistake in your css selector, you're using querySelector so inside the parenthese you're supposed to add the # if you're targetting an ID and a . if you're targetting a class. 
Your for loop is right but you need to update your variable everyProductsSubtotal for each iteration and set the totalPrice.innerHTML as everyProductsSubtotal. 

Based on what you've done here is a solution:

function calculateAll() {
  const allProducts = document.querySelectorAll(".product");
  everyProductsSubtotal = 0;

  for (let i = 0; i < allProducts.length; i++) {
    everyProductsSubtotal += updateSubtotal(allProducts[i]);
  }
  const totalPrice = document.querySelector("#total-value span");
  totalPrice.innerHTML = everyProductsSubtotal;
}

Iteration 4: Nice try but you're not targetting the right element, here your event element doesn't match with anything because you're not passing any argument when you're calling the removeProduct function line 59. To call properly the function and passing an argument to it you should use a forEach HOF to target each button: 

First thing your removeProductsBtn doesn't return anything (have you tried to console log it?) because your selector is not right, you're trying to select HTML tags which have a btn-remove class and which are inside another HTML tag with a btn class. You just want to target the elements with a btn-remove class. 
Then you need to pass as an argument the button you're clicking on to be able to remove the line (in other word the HTML tag tr with a class product associated to this button). 

Here is a solution for your event listener : 

let removeProductBtn = document.querySelectorAll(".btn-remove");
  removeProductBtn.forEach((btnRemove)=>{
    btnRemove.addEventListener("click", ()=> removeProduct(btnRemove));
  })
and here a proposal for your removeProduct function:

  const target = event.closest(".product").remove();
  console.log("The target in remove is:", target); 
}

Try to review your CSS selectors, it seems that you have some trouble to target the right element in your HTML file. Always console log the elements you're manipulationg so you know if you're on the right track ;) 

Wax on, wax off! :)

PR #1771, opened 2020-11-22
(PARIS FT 9/11) Jacqueline Jumaga

jjumaga:
I spent a lot of time on this and unfortunately couldn't get farther than this. I understand the concept well enough about needing to know how to access the correct element in the dom, but had problems find the right syntax (I think). I have some notes here as well as in my VS code to explain my thought process.

Iteration 2
- When I use getElementsByClassName(‘.product’) it returns an empty HTML collection.
- When I try to spread copy the HTML collection into an array it doesn’t work, I can’t figure out the proper syntax
- I tried using querySelectorAll(‘.product’) which give me an array with the 2 tr.product values, but when I try to use forEach to loop through the 2 product tr’s, the first product updates its subtotal but not the second, and the result in the console is to print NodeList(2) [tr.product, tr.product] twice

- You will find alternate syntax options that I tried that also didn’t work but I want to show you that I tried everything I could think of!

Iteration 3
I know I need to:
Access the total inside the span using DOM – I think I did this correctly because when I console.log it gives me <span>0</span>.

I need to add the result of adding all subtotals together and input this value into the total

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,7 +32,19 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
-        <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Beach Towel</span>
+          </td>
+          <td class="price">$<span>12.00</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
       </tbody>
       <tfoot>
         <!-- <tr class="create-product">
--- a/js/index.js
+++ b/js/index.js
@@ -1,26 +1,53 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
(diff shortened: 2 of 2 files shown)


VProum:
Hi Jacquie,
there is some concept that is not fully understand right here, but you have the good logic, keep practicing, it will come!

in the function updateSubtotal(product) from iteration 1
You are not using the "product" argument from the fonction

Line 4 and 6, you are doing "document.queryselector(...)" instead of "product.queryselector(...)"
this cause your function updateSubtotal to always select the first thing inside document
and that why the second product is not updating its subtotal

Also, there is no return value in your function updateSubtotal, so every calculation you have done is lost and you cannot use the function to calculate the total price

Line 46,
instead of totalValue += extractedSubtotal[i];
you should have done something like:  totalValue += updateSubtotal(Products[i]) (with a return value in the updateSubtotal)

 Side note :
you have two ways of selecting HTML from the JS
you can go with queryselector/queryselectorall OR with getElementsbyXXX
choose one or the other, but stick with it because the syntax is not the same :
queryselector expect a css selector inside (so with .class, #id etc..)
getElementsbyXXX expect a string, (so classe, id, etc WITHOUT the . # )

that is why "When you use getElementsByClassName(‘.product’) it returns an empty HTML collection."

PR #1845, opened 2021-02-12
[PAR-WFDT-102] Benjamin PINARD

benpin31:


ta-web-paris:
Nice job on this one Benjamin ! 
It's true that the createRow was tricky, because it can reset things if you do it the wrong way

But you figured it out :)

PR #1846, opened 2021-02-12
[PAR-WDFT-102] Romain Benard

romainb33:
Please find my LAB,
have a nice week-end

### DIFF
--- a/index.html
+++ b/index.html
@@ -48,7 +48,7 @@
         </tr>
       </tbody>
       <tfoot>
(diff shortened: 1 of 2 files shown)


VProum:
Hi Romain!

Excellent job, everything is working fine and you did all the bonuses!
Nice use of HOF
instead of using parentnode.parentnode, you could have use closest()
https://developer.mozilla.org/fr/docs/Web/API/Element/closest

but as long as it works... :D

PR #1847, opened 2021-02-12
[PAR-WDFT-102] Julie Plantey

ronronscelestes:
Thanks to Vincent (& Andy) my createProduct function is not 10 pages long 🥳


ta-web-paris:
Nice job ! Very clever to check that the product has a name and price before creating the row :)

PR #1848, opened 2021-02-12
[WEB-DEV-PARFT] Giorgia Galasso

giorgiagalasso:


VProum:
Hi Giorgia, 

your querySelector are correct, 
you will have to use the updateSubtotal variable if you want so see some modification on your page

also try to console log price.innerHTML and quantity.innerHTML to see what is the const subtotal!

you can check the reinforcement of the 17th of febuary where i reviewed the lab including the bonus
feel free to come and see us if there is concept you didn't understand!

PR #1849, opened 2021-02-12
[PAR- WDFT-102] Thaïs HAUGUEL

thaishauguel:


VProum:
Hi Thaïs!

Excellent job, everything is working fine and you did all the bonuses!
Your querySelector line 41 are a bit strange but as long as it works :D
using nth child would be prettier
and maybe try to use HOF like foreach instead of a for loop?

PR #1850, opened 2021-02-12
[PAR-WDFT-102] NAOMI YAMANAKA

Naomax703:


ta-web-paris:
Hey Naomi !
Good job on the first iterations, one small thing : you're not inserting the value of the total in the HTML, you need to add something so the value goes inside ;)

Did you attend the reinforcement with Vincent ? I hope it was helpful ! You also have the solution code in the cohort folder, if you have any questions please feel free to ask us :)

Nina

PR #1852, opened 2021-02-12
[PAR-WDFT-102] Magali Fort

mflynx:
Have a great weekend


ta-web-paris:
Nice job on this one !
It looks good, just one quick thing : you're repeating the "listenRemove" function (even if you don't have errors) i'm sure you can have less calls of this one to have the same result :)

PR #1853, opened 2021-02-12
[PAR-WD-102] Anaïs Engler

anachannn:
Best lab so far!! I had fun doing it :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,10 +32,22 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
-        <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Beach Towel</span>
+          </td>
+          <td class="price">$<span>12.50</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
       </tbody>
       <tfoot>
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Anaïs, thanks for submitting your lab. 

Good job overall, just a few comments:

For iteration 4, try to have a look at closest method it will be way easier to target an element than using parentNode a bunch of time. But the way you did it is working so it's ok. 

For iteration 5, you'll notice that when you add a product all the inputs of the other products are removed. It's because you're adding a line directly inside the tbody with an innerHTML, you should rather create an tr element and add the innerHTML on it and add it to the tbody with an appenChild method. 

function createProduct() {

  const newProductName = document.querySelector(".create-product :nth-child(1) input").value;
  const newProductPrice = document.querySelector(".create-product :nth-child(2) input").value;
  
  let productCart = document.querySelector("#cart tbody");

  const newRow = document.createElement("tr");

  newRow.className = "product";

  newRow.innerHTML += `

 <td class="name"><span>${newProductName}</span></td>

 <td class="quantity">

 </td>

 <td class="action">

 </td>`;

 productCart.appendChild(newRow);



  //Two following lines possible with productCart.appendChild

  // const lastBtn = productCart.querySelector(".product:last-of-type").querySelector(".btn-remove");

  //lastBtn.onclick = removeProduct;

  const removeBtnList = document.querySelectorAll(".action .btn-remove");

removeBtnList.forEach(removeBtn => removeBtn.addEventListener('click', removeProduct));

}

Keep in mind that you can target the last button added with the commented lines I add in the function instead of using a forEach HOF and going through all the remove buttons again. 

If you have any questions, don't hesitate! Good job!

Wax on, wax off! 

Fanny

PR #1854, opened 2021-02-12
[PAR-WDFT-102]bouba Sidibe

armarmlegs:
done up to iteration 3, bonus for the weekend

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ We already declared `removeProduct(event)` and added some starter code. After yo
 
 As we can see, `removeProduct(event)` expects the event as a single argument, and that is going to trigger the removal of the corresponding product from the cart.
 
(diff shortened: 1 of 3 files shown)


Fanny-Sbn:
Hello Bouba, thanks for submitting your lab. 

Good job for the mandatory iterations, here are a few comments: 

Iteration1: Cool

Iteration 2&3:  You could have use a HOF but it's working perfectly with a for loop :)  You can optimize it a bit, you should target and update your total in your HTML outside of your for loop, right now you're retargetting and updating "superduperkyle" (hahahahaha) on each turn... 

Iteration 4: Nice but try to have a look to the closest method, using parentNode a bunch of times is not really elegant. Don't forget to call your calculateAll function inside your removeProduct function so that your total can be updated consequently. 

Try to do the last bonus to practice DOM, nice job overall!

Wax on, wax off! 

Fanny

armarmlegs:
Thank you fanny!

It's noted :) will give it a try

Le lun. 22 févr. 2021 à 12:33, Fanny-Sbn <notifications@github.com> a
écrit :

PR #1856, opened 2021-02-12
[PAR-WDFT-102] Anna Lefour

AnkaPieka:
Hi,
not sure my code is DRY but it works.
See you

### DIFF
--- a/index.html
+++ b/index.html
@@ -35,7 +35,7 @@
 
         <tr class="product">
           <td class="name">
-            <span>Ironhack Rubber Cup</span>
+            <span>Ironhack Cup</span>
           </td>
           <td class="price">$<span>10.00</span></td>
           <td class="quantity">
--- a/index.html
+++ b/index.html
@@ -49,7 +49,7 @@
         
       </tbody>
       <tfoot>
-        <!-- <tr class="create-product">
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Anna, thanks for submitting your lab. 

Here are a few comments:

Iteration 1 : It's ok but prefer to change the value in your HTML file with DOM manipulation for your result and then return only the subtotal, not "result.innerHTML = subtotal, it's weird. 

Iteration 2 & 3: It's working but unfortunately it's hardcoded and your function won't be reusable when you'll add other products. 

Here you need to use your updateSubtotal function to be able to calculate each product separately and make the sum of all this product to have the total price. To do so you can simply create a totalPrice variable and add to it the result of the updatesubTotal function for each product. You were on the right path with : 

products.forEach(product => {
    updateSubtotal(product);
  }); 

You just need to adapt it a bit (here is the function calculateAll)

function calculateAll() {
 let products = document.querySelectorAll('.product');
let TotalPrice =0;

products.forEach(product => {
    totalPrice += updateSubtotal(product);
  }); 

let totalOfCart = document.querySelector("#total-value span");
totalOfCart.innerHTML = totalPrice;

PR #1859, opened 2021-02-13
[PAR-WDFT-102] Myriam MIRA

myracodes:


ta-web-paris:
Hi Myriam ! is it working properly ? Because for instance line 13, you're not passing the product as a parameter to the function updateSubtotal, so i'm not sure it works like that. I see your push was before the reinforcement, have you been to the reinf ? If so, did it help you understand the rest of the lab ?

Please tell me how you feel about it now, because after a week of project and the reinforcement i don't want to spend too much time if you understood the concepts afterwards :)

Nina

PR #1861, opened 2021-02-13
[PAR-WDFT-102] Pauline Bellaud

Iron-Popi:
Tried to use a loop for iteration 3, but kept getting NaN as a result :(
So I did it this way, not the most practical but hey it's working

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,6 +33,32 @@
           </td>
         </tr>
         <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Unicorn Plush</span>
+          </td>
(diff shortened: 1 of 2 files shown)


VProum:
Hi Pauline,

Nice work on the first 2 iterations,
for the 3rd one :
the code you have made will only work with those 3 products because it is hard coded, if you add a line or remove one it will not work anymore

didn't you feel a repetition coding this? you could have use for loops!

```
const Allnumb = parseFloat(document.querySelectorAll('.subtotal span'));
let total = 0;
for ( let i=0; i<Allnumb.lenght; i++){
   total = total + parseFloat(document.querySelectorAll('.subtotal span')[i].innerHTML
}
```
hope this helps

PR #1864, opened 2021-02-15
[PAR-WDFT-102] YUKA TANAKA

seieeeen:
-Iteration 2 done, the rest was quite hard...

### DIFF
--- a/index.html
+++ b/index.html
@@ -27,13 +27,26 @@
           <td class="quantity">
             <input type="number" value="0" min="0" placeholder="Quantity" />
           </td>
-          <td class="subtotal">$<span>0</span></td>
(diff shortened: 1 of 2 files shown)


VProum:
Hi Yuka,

Nice work on the first 2 iterations, 
for the 3rd one, you will need to select you span with a sumSubtotal class, and put the result of the sum in the innerHTML

you will need to do that inside your calculateAll() function because you only want to calculate the total price after a user has click on the button, and calculateAll is the handler triggered when you click on it!

keep in mind that getElementsByClassName return an array, so you can't do innerHTML on a array
i would suggest you stick with queryselector if you want to select one html element and queryselectorAll if you want to select multiple html element

PR #1866, opened 2021-02-16
[PAR-WDFT-102] Fatou D

Fatoumata-lab:


Fanny-Sbn:
Hello Fatoumata, thanks for submitting your lab. 

Here are a few comments: 

First of all, always check that you don't have syntax error before submitting your lab, with syntax errors your code won't work. 
When I'm checking on the browser I can only see one product but normally if you read carefully the instructions, they ask you to add another one directly in the HTML file, so pay attention to the instructions. 

Iteration 1: You have the right logic but you're missing two steps -> it's better to store the subtotal in a variable, so create one variable subtotal where you'll store the product of the price and the quantity. 
Then you'll need the value of that this function will return, so you need to **RETURN** something (here the subtotal). 

Iteration 2&3:  You need to use the first iteration, the updateSubtotal function to calculate the total, you have the logic but you're not using the right elements. 
Basically you declare a variable sumOfSubtotals, that's great. What does it represent? the sum of all subtotals, ok. How can you have the subtotals? Huuum, we have a function updateSubtotal that returns the subtotal of a product so we need to use it to calculate the sum of subtotals. 
For each product we will add to the subtotal of this product to the sumOfSubtotal

function calculateAll() {
  let sumOfSubtotals = 0;
  document.querySelectorAll('.product').forEach(product=> {
    sumOfSubtotals += updateSubtotal(product);
  }); 
  document.querySelector('#total-value span').innerHTML =sumOfSubtotals;
}

If you have any questions, don't hesitate to reach us out. 

Wax on, wax off! 

Fanny

PR #1868, opened 2021-02-16
[PAR-WDFT-102] Alexis Verthuy

AlexisVerthuy:
Only Iteration1 done

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,6 +32,19 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
+          <td class="name">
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Alexis, thanks for submitting your lab. 

For the first iteration, you're missing one thing! Keep in mind that you'll need to use the value that this function will **return**, so you aboslutely need to return something or you'll will have "undefined"!

Have you tried to do the rest of the iterations or were you really stuck from the first iteration?
Don't hesitate to reach us out if you need to review some of the concepts or if you have any specific questions. 

Wax on, wax off!

Fanny

PR #1873, opened 2021-02-22
[PAR-WDFT-102] Augustin LEMAN

august159:
Forgot to pull request last week

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,9 +33,35 @@
           </td>
         </tr>
         <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack foam</span>
+          </td>
+          <td class="price">$<span>8.00</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Augustin, thanks for submitting your lab. 

Here are a few comments: 

Iteration 1: Perfect

Iteration 2 & 3: You totally have the great logic but you made a small mistake and your total are not updated in your browser. You're misusing innerText! innerText is not a function, it's a property. 

Iteration 4: 
You have several mistakes.

- Have a look to the removeChild function on mdn and look at what this function is expecting. Here you need to pass as an argument parentTarget and not target. 
Using parentNode a bunch of time is not so elegant, you can have a look to the closest method instead. 

- You have some mistakes when doing your event listener, try to console log removeBtns and see what you get ( you'll be surprised), have a look to your selector. Plus, the addEventListener function is expecting as a second argument a callback function, here you're running directly your function, you must remove the parentheses. 

If you have any questions, don't hesitate to reach us out. 

Wax on, wax off! 

Fanny

PR #1893, opened 2021-03-30
BER-WDFT-MAR21 Markus Heisinger, Jan Colomer

mhsngr:


js/index.js
```diff
@@ -2,41 +2,75 @@
...
+  <td class="price">$<span>${price}</span></td>
+  <td class="quantity">
+  <input type="number" value="0" min="0" placeholder="Quantity" />
+  </td>
+  <td class="subtotal">$<span>0</span></td>
+  <td class="action">
+  <button class="btn btn-remove">Remove</button>
+  </td>`;
+  document.querySelector('tbody').appendChild(newProduct);
 }
```
ahmedhany82, line 62: I think you missed adding an event listener for the remove button in the new created product. Currently the remove button does not work for new products. May be somthing like so:

`newProduct.querySelector('.btn-remove').addEventListener('click', removeProduct);`

I would also reset the input fields after appending the new product.
```
  document.querySelector('.create-product input[type="text"]').value = '';
  document.querySelector('.create-product input[type="number"]').value = 0;
```

PR #1896, opened 2021-03-30
[Berlin-WDFT-MAR2021] Marcia Faria Netter

mfndnc:


js/index.js
```diff
@@ -1,23 +1,40 @@
...
   // ITERATION 3
   //... your code goes here
+  const totalPricesDom = document.querySelectorAll('.product .subtotal span');
+  let totalPrice = 0;
+  for (let domEl of totalPricesDom) {
+    totalPrice += Number.parseFloat(domEl.innerText);
+  }
+  document.querySelector('#total-value span').innerText = totalPrice.toFixed(2);
+  //console.log(document.querySelector('.product .subtotal span').innerText);
+  // console.log('totalPrice', totalPrice);
```
ahmedhany82, line 37: Good Job! Another way to do it is to make `updateSubtotal` function return the subtotal and then you just accumulate it in the loop above in line 25.

js/index.js
```diff
@@ -26,17 +43,56 @@ function removeProduct(event) {
...
+            <td class="name"><span></span></td>
+            <td class="price">$<span></span></td>
+            <td class="quantity"><input type="number" value="0" min="0" placeholder="Quantity" /></td>
+            <td class="subtotal">$<span>0</span></td>
+            <td class="action"><button>Remove</button></td>`;
+  tr.querySelector('button').classList.add('btn', 'btn-remove');
+  tr.querySelector('button').addEventListener('click', removeProduct);
+
+  tr.querySelector('.name span').innerText = productNameDom.value;
+  tr.querySelector('.price span').innerText = productValDom.value;
```
ahmedhany82, line 78: Great job for this one as well. Just to offer you an additional way to do it. You can fist define two variables somewhere before line 68:
```
  let newProductName = productNameDom.value;
  let newProductPrice = productValDom.value;
```
And then when creating the innerHTML, you refer to them as variables:
```
  tr.innerHTML = `
            <td class="name"><span>${newProductName}</span></td>
            <td class="quantity"><input type="number" value="0" min="0" placeholder="Quantity" /></td>
            <td class="action"><button>Remove</button></td>`;
```

ahmedhany82 (commented):
Hey Marcia, good job for this lab! I only added 2 small comments offering an alternative way to do things. Keep up the good work! Regards!

PR #1898, opened 2021-03-30
BER WDFT MAR2021 - Olesia, Diogo

olakul:


annawaszak (commented):
Hi Ola and Diogo,

You have done a good job on the lab so far. There really isn't anything you can improve here. If you decide to go for iterations 3 - 5 and would like my feedback, give me a shout and I'll take a look again!

Thanks,
Anna

PR #1899, opened 2021-03-30
[BER-FT-032021] Philipp Dam

Philipphub:


js/index.js
```diff
@@ -1,31 +1,37 @@
...
+  let sum = [];
+  products.forEach((item) => {
+    sum.push(updateSubtotal(item));
+  });
 
   // ITERATION 3
-  //... your code goes here
+  document.querySelector("#total-value span").innerHTML = sum.reduce(
+    (a, b) => a + b
+  );
```
annawaszak, line 23: If you made the sum a number, instead of an array, you could save yourself some time on the reduce.

annawaszak (commented):
Hi Philipp, 

Excellent job on this one. Looks like you can do manipulate DOM in your sleep :)
The only suggestion I could make is that sometimes your code is so efficient that it becomes less readable.

Thanks,
Anna

annawaszak:
If you decide to do the last iteration and want some feedback, just let me know!

PR #1900, opened 2021-03-30
[BER-WDFT-032021] Alexander Rathbone - Pietro Ballarin

pietroballarin:
first draft

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -18,8 +18,6 @@ let sumOfSubtotals = 0;
 for (let i = 0; i < productToPass.length; i++) {
 sumOfSubtotals += updateSubtotal(productToPass[i])
 totalValue.innerHTML = sumOfSubtotals;
-
-console.log(totalValue)
 }
 
  /* const singleProduct = document.querySelector('.product');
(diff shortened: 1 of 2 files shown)


pietroballarin:
Almost done. Remove button and calculate price functionalities for new items are missing

js/index.js
```diff
@@ -24,19 +35,66 @@ function calculateAll() {
...
+  nameValue = ''
+  priceValue = ''
+  
 }
 
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
 
+  const removeBtn = document.querySelectorAll('.btn.btn-remove');
```
annawaszak, line 90: Since all the remove buttons are assigned to this variable, I would recommend using a plural name.

js/index.js
```diff
@@ -24,19 +35,66 @@ function calculateAll() {
...
 }
+testRow()*/
 
 // ITERATION 5
 
-function createProduct() {
-  //... your code goes here
+function createProduct(event) {
+  const targetName = document.querySelector('.create-product').querySelector('input[type="text"]')
+  let nameValue = targetName.value;
```
annawaszak, line 55: i would have probably suggest to add a class to that field so that you can access it more easily.
But if you would like to leave HTML as it is, then instead of chaining querySelectors you could access it like this:
`  const targetName = document.querySelector('.create-product > td > input[type=text]').value`

annawaszak (commented):
Hi Alex and Pietro,

You have done a really good job on the lab! Congrats on making it through all the iterations.

Below you can find a few comments or suggestions. If you have any questions, just leave a comment and I'll get back to you asap.

Thanks,
Anna

PR #1903, opened 2021-03-30
done

Raffaeleb003:


js/index.js
```diff
@@ -2,41 +2,85 @@
...
   // ITERATION 3
-  //... your code goes here
+  const eachTotal = document.querySelectorAll('.subtotal span');
+  let allTotal = 0;
+
+  for (let i = 0; i < eachTotal.length; i++) {
+    allTotal += parseInt(eachTotal[i].innerHTML);
+  }
+  document.querySelector('#total-value span').innerHTML = allTotal;
 }
```
ahmedhany82, line 43: Good job for this one. Another way to do it is use the return from `updateSubtotal` to calculate the total sum. See below:

```
  let totalPrice = 0;
  products.forEach(function(product) { 
    totalPrice += updateSubtotal(product);
  })
```

js/index.js
```diff
@@ -2,41 +2,85 @@
...
 // ITERATION 4
 
 function removeProduct(event) {
   const target = event.currentTarget;
   console.log('The target in remove is:', target);
-  //... your code goes here
+  const productToRemove = event.currentTarget.parentNode.parentNode
+  // document.productToRemove('.cart').removeChild(productToRemove)
+  productToRemove.remove()
 }
```
ahmedhany82, line 54: Good job for this one! What is missing here is to update the total price after removing products. You do that by just calling `calculateAll()` at the end of the function to recalculate the total.

js/index.js
```diff
@@ -2,41 +2,85 @@
...
+  newProduct.querySelector('.product .name').innerText = newProductName
+  newProduct.querySelector('.product .price').value= newProductPrice
+
+  document.querySelector('#cart tbody').appendChild(newProduct)
 }
 
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
+  calculatePricesBtn.addEventListener('click', calculateAll);
```
ahmedhany82, line 75: This line is repeated I guess.

js/index.js
```diff
@@ -2,41 +2,85 @@
...
+  productToRemove.remove()
 }
-
 // ITERATION 5
-
 function createProduct() {
   //... your code goes here
+  let newProductName = document.querySelector('.create-product .name').innerContent;
+  console.log(newProductName)
+  let newProductPrice = document.querySelector('.create-product .price').value;
```
ahmedhany82, line 60: I think there are no classes named `name` and `price` inside the `create-product `class in the html. You could probably use something like `const inputValues = document.querySelector('.create-product').querySelectorAll('input')` to get all the input fields. Then `inputValuess[0].value` will get you the name and `inputValuess[1].value` will get you price.

js/index.js
```diff
@@ -2,41 +2,85 @@
...
+  console.log(newProductPrice)
+
+  const referenceProduct = document.querySelector('.product');
+  let newProduct = referenceProduct.cloneNode(true)
```
ahmedhany82, line 64: This is a good idea to clone but what if you removed all the products and then you want to create a new product. One way to approach this is to create an html template for a product as a string, insert a row in the table using `insertRow()`, then inject your template into this new row using `innerHTML `. Give it a thought and we can discuss it.

js/index.js
```diff
@@ -2,41 +2,85 @@
...
+  calculatePricesBtn.addEventListener('click', calculateAll); 
 
   //... your code goes here
+  const removeBtn = document.querySelectorAll('.btn-remove');
+  removeBtn.forEach(eachRemoveBtn => {
+  eachRemoveBtn.addEventListener('click', removeProduct);
+})
+
+const createProductBtn = document.querySelector('#create');
+createProductBtn.addEventListener('click', createProduct);
```
ahmedhany82, line 84: This is causing an error: `Uncaught TypeError: Cannot read property 'addEventListener' of null` mainly because you commented out the create product part in the html.

ahmedhany82 (commented):
Hey Raffaele, good work for this lab. Please find my review comments. I encourage you to give the `creatProduct()` a second try. Feel free to reach out in case you have any questions.

PR #1915, opened 2021-03-30
[FTAMS032021] Jeroen

jeroenvanbart:
done exept for bonus

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,23 +1,25 @@
 // ITERATION 1
-
+ 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  const price = product.querySelector(".price span").innerHTML; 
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,23 +1,25 @@
...
-  updateSubtotal(singleProduct);
-  // end of test
 
-  // ITERATION 2
-  //... your code goes here
 
-  // ITERATION 3
-  //... your code goes here
+function calculateAll() {
+  const products = [...document.getElementsByClassName('product')];
```
reMRKableDev, line 15: Good use of the spread operator here `...`

reMRKableDev (commented):
Nice work @jeroenvanbart 🚀 

The one thing I found missing is the creation of a new product for iteration 2. Otherwise, there is only one product for which the total is calculated for. Your logic and implementation will still work in that case.

PR #1916, opened 2021-03-31
[BER-FT-30-03-2021] Completed DOM Ironhack Cart Assignment | Chiara, Cintia

chiarawilden:
Finished Iteration 1-3!

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -2,22 +2,31 @@
 
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  let price = 0;
+  let quantity = 0;
+  price = product.querySelector('.price span').innerText;
+  quantity = product.querySelector('.quantity input').value;
+  let subtotal = price * quantity;
+  product.querySelector('.subtotal span').innerText = subtotal;
+  return subtotal;
 }
 
+// ITERATION 2 + ITERATION 3
+
+// code in the following two lines is added just for testing purposes.
+// it runs when only iteration 1 is completed. at later point, it can be removed.
+// const singleProduct = document.querySelector('.product');
+// updateSubtotal(singleProduct);
+
 function calculateAll() {
-  // code in the following two lines is added just for testing purposes.
-  // it runs when only iteration 1 is completed. at later point, it can be removed.
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
-  // end of test
-
-  // ITERATION 2
-  //... your code goes here
-
-  // ITERATION 3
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -2,22 +2,31 @@
 
 function updateSubtotal(product) {
```
wax-bear, line 3: A bit of vertical spacing would improve the readability. For example like this:

` let price = 0;
  let quantity = 0;

  price = product.querySelector('.price span').innerText;
  quantity = product.querySelector('.quantity input').value;

  let subtotal = price * quantity;

  product.querySelector('.subtotal span').innerText = subtotal;

  return subtotal;`

js/index.js
```diff
@@ -2,22 +2,31 @@
...
   console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  let price = 0;
```
wax-bear, line 5: How about declaring price and assigning the innerText of the query selectors ".price span" to it? In that case, could a const be used instead of a let?

js/index.js
```diff
@@ -2,22 +2,31 @@
...
+  let quantity = 0;
+  price = product.querySelector('.price span').innerText;
+  quantity = product.querySelector('.quantity input').value;
+  let subtotal = price * quantity;
```
wax-bear, line 9: Will this variable be reassigned? If not, how about using const?

js/index.js
```diff
@@ -2,22 +2,31 @@
...
+// code in the following two lines is added just for testing purposes.
+// it runs when only iteration 1 is completed. at later point, it can be removed.
+// const singleProduct = document.querySelector('.product');
+// updateSubtotal(singleProduct);
 
-  // ITERATION 3
-  //... your code goes here
+function calculateAll() {
+  let totalSum = 0;
+  total = document.querySelector('#total-value span');
```
wax-bear, line 23: A variable keyword (let/const/var) is missing here. This way, total will be globally available when calculateAll is called.

js/index.js
```diff
@@ -2,22 +2,31 @@
...
+  const products = document.getElementsByClassName('product');
+  for (let i = 0; i < products.length; i++) {
+  //updateSubtotal(products[i]);
+  totalSum += updateSubtotal(products[i]);
```
wax-bear, line 27: - A tab is missing here. The entirety of the code block should be indented.
- totalSum is a global variable. Please make it function/block scoped.

wax-bear (commented):
Great stuff Chiara and Cintia!
Please check out the comments for potential improvements.
Please reach out if you have a question.

Best,

Thomas

PR #1937, opened 2021-04-04
[FTAMS032021] Igor Godinho Figueiredo

Igor-GF:
I've got confused in the bonus regarding the DOM manipulation with HTML collections / elements.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,9 +1,11 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  let price = document.querySelector('.price span').innerHTML();
+  let quantity = document.querySelector('.quantity input');
+  let subTotal = document.querySelector('.subtotal span').innerHTML();
+  subTotal.innerHTML = price * quantity;
+  return subTotal;
 }
 
 function calculateAll() {


js/index.js
```diff
@@ -1,9 +1,11 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  let price = document.querySelector('.price span').innerHTML();
```
reMRKableDev, line 4: Ok, I see where you are trying to go with this. Keep in mind, however, that `innerHTML` is an attribute in the DOM element and not a function. Calling `innerHTML()` results in a TypeError to be thrown in upon running this. 

To get your updateSubtotal working, consider having the following steps in your logic:

1. Get DOM elements that hold price and quantity. 
2. Extract values from DOM elements.
3. Calculate total values.
4. Get DOM element that holds the subtotal value for the product.
5. Set the product subtotal to the corresponding DOM element.
6. Return subtotal value so it can be used later.

reMRKableDev (commented):
Good attempt @Igor-GF 🚀 

Your `updateSubtotal()` solution isn't completely done. I've left some comments on how you can make it work out in the end. Check it out and apply the changes

PR #2000, opened 2021-06-08
[BER-WDFT-MAY2021] Alexandre Oliveira

alexandreoliv:
Some tests are failing, even though the functions are working. I might need to fix something.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,42 +1,78 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  const price = product.querySelector('.price span').textContent;
+  const quantity = product.querySelector('.quantity input').value;
+  const subtotal = product.querySelector('.subtotal span');
+  
+  subtotal.textContent = price * quantity; // Updates subtotal on the page
+  return Number(subtotal.textContent);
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -1,42 +1,78 @@
...
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
+  `;
+
+  parent[0].innerHTML += newProduct; // Adds a new product to the table
```
ahmedhany82, line 56: A small remark here. When you add a new product, the quantity of the existing products goes back to 0. I think because the html has 0 as the quantity of the products and in this line it just replaces what is in the DOM. You could also use `insertRow()`.
```
  let newRow = parent[0].insertRow()
  newRow.className = 'product'
  newRow.innerHTML = newProduct;
```
This will keep the remaining products unchanged.

ahmedhany82 (commented):
Hey Alexandre,
Great job for the lab. I wouldn't worry too much about the test cases for this lab. I can see the functionality works well. I just added one comment for adding a new product.
Best regards,
Ahmed

PR #2011, opened 2021-06-08
Bart Spangenberg: Lab Shopping Cart

BartSpangenberg:
Hey TA's,

Test cases are not working but I think everything works as it should.

Cheers!

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,10 +32,36 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+        <tr class="product">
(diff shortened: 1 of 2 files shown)


TA-Remote:
Wow Bart! Really well done. You did everything and you even went a step further with the alert. Awesome 💯 A couple of things to improve: get rid of useless console.logs when you are done and for the create button you could have used string interpolation and it should've work too but with 50 lines of code less. 👍 Great Job!

PR #2017, opened 2021-06-08
Ellen Children - WDFT - RMT - MAY21

elyzx:
Hey TAs! 

This was a tough one. The tests are failing but I'm unsure how to fix. Curious to see the solution ;)


TA-Remote:
Hi Ellen! Your logic is good and I think you would have passed the tests if you understood better the instructions. You went for document instead of product. But your logic is definitely good! Keep going like that!

PR #2019, opened 2021-06-08
Joanne Seiler WDFT-RMT-MAY21

Joanneseiler:


TA-Remote:
Nice job on this lab and the bonus :) Feel free to check the solution to see how you could have done bonus 2

PR #2025, opened 2021-06-09
George WDFT-RMT-MAY21

gdsbrooks:
Challenging, tho started to make more sense. I created a document fragment to duplicate and append the newly created Item. If all Items are removed, create cannot add a new one. I tried to initialise / create the document variable before any remove functions were called, or store a copy of the document fragment in a global variable but this did not resolve it. Hmmmm.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -43,7 +43,7 @@ table input[type='number'] {
 .btn {
   display: inline-block;
   border: 0;
-  border-radius: 0.25em;
+  border-radius: 0.5em;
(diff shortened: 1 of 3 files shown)


TA-Remote:
Well done George. 💯  You did all the iterations including bonus, awesome!. 🏅 Remember you have to call the functions inside the window load event listener, that way any of the magic will happend untill the whole document is loaded and you'll prevent for the future posible bugs and errors. 👍 Good job! Keep it up!

PR #2027, opened 2021-06-09
KALANDE-RMT-MAY21-WDFT

Kalande:
Planned to do iteration 5 this morning, woke up too late :(


TA-Remote:
Hey Kalande! Very good job, you pass all the tests. Give a try for iteration 5 you can do it easily I think even if the response is longer. Keep going!

PR #2028, opened 2021-06-09
Anja WDFT-RMT-MAY21 Ironhack cart

AnjaRL:
solved lab excluding iteration5
Very very very tough one... needed more than one night to do it 😞🧐


TA-Remote:
Very nice job on the lab, doing at least one bonus is really great. Don't worry about the last bonus, just check the solution to see how it could have been done :)

PR #2057, opened 2021-06-11
Update test specs

ross-u:
- Add an additional test spec
- Update "should allow creation of product" spec
- Update the setup of all `describe` blocks
- Update the setup for "Remove products" and "Create products" `describe` blocks to remove existing products before arranging and assertion.


sandrabosk:
Thank you @ross-u, looks good.

PR #2084, opened 2021-07-15
[FT-RMT-072021] KarinaNarkeviciute

Karina-N:


Karina-N:
Still some bugs left:
--> remove button does not work of newly created items
--> if I remove all original products, clone method does not work anymore so I cannot create new products at all

Karina-N:
Fixed remove button, but now the  total sum is being calculated without clicking calculate button...

PR #2206, opened 2021-10-26
Berlin Noam and Johannes

johann1301:
changes


Farxa:
Hi Noam and Johannes,
I checked your solution. Good job, I honestly have no other feedback on it. It is good to remove the console.log after you are done to keep the code nice and clean, and you can always come back and do the bonuses when you have time and want to practice.

Cheers, Rahaf

PR #2223, opened 2021-10-26
full solution for table manipulation / @mahsa

mahrad92:


Farxa:
Hi Mahsa,
I checked your lab solution, good job. I have nothing to add 👍🏽 but try to remove all the console logs after you are done with the lab, this makes our code cleaner for better readability.

PR #2272, opened 2021-11-28
[MAD PT 112021] María Gallego

MaGaBoo:


MaGaBoo:
Currently trying to get Total price. Main problem is allSubtotals variable. I try to convert it into an array, but I get... two different arrays.

MaGaBoo:
Still trying to solve iteration 2, I´m totally stuck on this. Started from the beginning and made changes; now I get "null" instead of "undefined" on total $.
Really need help with this :-/

PR #2348, opened 2022-01-18
Areeba-BER-FT-Jan'22  Done with lab including bonus

Areeba-Imran:


stephen-gallagher:
Nice work Areeba!! 

Just a little tip that could simplify the createProduct() function - instead of inserting every cell individually, you can actually just create a new 'tr' and then paste the whole block of code for the product in your html file as the innerHTML of the new 'tr' you created. It could look something like this:

`let newRow = table.insertRow();
  newRow.classList.add('product');
  newRow.innerHTML = '
    <td class="name">
      <span>${productName}</span>
    </td>
    <td class="price">$<span>${productPrice}</span></td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>';`

PR #2356, opened 2022-01-18
GJ & Aleksandra lab cart

Aleksandrawoslx:


js/index.js
```diff
@@ -1,18 +1,35 @@
...
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector('.quantity input');
+  const priceValue = price.innerHTML;
+  const quantityValue = quantity.value;
```
TA-Remote, line 8: Here you can simplify code, to make it even shorter. By inserting line 5 code directly instead of word 'price' on line 7. So below syntax is equivalent to what you have:

```
  const priceValue = product.querySelector('.price span').innerHTML;
  const quantityValue = product.querySelector('.quantity input').value;
```

js/index.js
```diff
@@ -1,18 +1,35 @@
...
+  const subTotalPrice = priceValue * quantityValue;
+  const subTotal = product.querySelector('.subtotal span');
+  const subTotalCalc = subTotal.innerHTML = subTotalPrice;
```
TA-Remote, line 11: If you will not be using this variable again, do not need to declare it. You can simply write the functionality -> find the element and update the innerHTML of it.

```
product.querySelector('.subtotal span').innerHTML = subTotalPrice;
```

js/index.js
```diff
@@ -1,18 +1,35 @@
...
-  updateSubtotal(singleProduct);
-  // end of test
-
+  let allProducts = document.querySelectorAll('.product');
+  allProducts = [...allProducts];
+
+  let totalValue = 0;
+  allProducts.forEach(function (element) {
+     updateSubtotal(element);
+     totalValue += updateSubtotal(element)
```
TA-Remote, line 26: line 25 is unnesessary, since we are already getting the result of this directly on line 26.

PR #2359, opened 2022-01-18
Olayinka WDFT BER22

ola-kiitan:


stephen-gallagher:
Nailed it Olayinka! Everything looks great - just a couple things to simplify the code a little bit:

1. You don't need to put any parameters in the brackets when using removeChild()
2. You don't need to put the removeButtons logic inside of a forEach() inside of your createProduct function, as there will only be one remove button in every product that you create

ola-kiitan:
HI Stephen,

Thank you for the feedback. I will pay more attention to that next time.

PR #2361, opened 2022-01-18
Lab solved

Jan-000:
This is first push of Iron-cart Lab.

Got stuck with the last bonus;
I heavily relied on TAs and githubs of others to get it done.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,22 +32,38 @@
             <button class="btn btn-remove">Remove</button>
           </td>
         </tr>
+
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Iron</span>
+          </td>
+          <td class="price">$<span>10.00</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
+
         <!-- Iteration 2: Add more products here -->
       </tbody>
       <tfoot>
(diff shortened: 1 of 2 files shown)


stephen-gallagher:
Nice work Jan! Don't discredit your results due to receiving guidance from others - the main purpose of these labs is for you to understand the key concepts that are outlined in the iterations. As long as you have a good understanding as to why each line of code is written and how it is implemented then it's totally fine to receive assistance - it's what we're here for!

Just in regards to the last iteration with the remove buttons - if you scroll down to the bottom of your page (line 78) you can find the code that you have implemented for the remove buttons. We can see that this logic is nested inside of the following function: 

`window.addEventListener('load', () => {`

the 'load' keyword is important in this case - this basically means that all the code that is written inside of this event listener is only going to be applied to the elements that are already on the page when the page is loaded (e.g when you refresh the browser). This means that your remove button logic will only be functional for the 2 existing products on the page. As we are then filling out a form to create a new product, the logic will not be accessible for this new remove button as the page has already completely loaded. This issue can be resolved in a couple of ways:

1. You could add a new event listener inside of your createProduct() function that will listen for the 'click' of this new remove button that has been created that will call the removeProduct() function
2. You could bring your remove button logic (lines 78 - 81) into its own function (e.g removeItem() ), and then call that function both in the 'load' event listener at the bottom of the page, and the createProduct() function when you add a new item


Sorry for the huge block of text i just wrote but I hope that helps clear things up a bit!

Jan-000:
Hey Stephen, I appreciate such a clear answer!
Thanks for that : ) I made a step forward thanks to this.

PR #2363, opened 2022-01-18
[Remote FT 012022] Denise Sengül

denisesenguel:


js/index.js
```diff
@@ -3,6 +3,14 @@
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
 
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector(".quantity>input");
+  const priceValue = parseFloat(price.innerHTML);
+  const quantityValue = parseInt(quantity.value);
```
TA-Remote, line 8: Here you can combine lines to write shorter code - as long as it is clear to you. So lines below are identical to lines 6-9.

```
  const price = parseFloat(product.querySelector('.price span').innerHTML);
  const quantity = parseFloat(product.querySelector('.quantity input').value);
```

PR #2365, opened 2022-01-18
[BER WD FT JAN 2022] Joana Carballo

jocarballo:


stephen-gallagher:
Looks good Joana, every iteration looks great! Just a couple of tips to help simplify your code a little bit: 

**1.** **remove product**
The 'closest' command works great here, but if you want to implement this logic in a couple of lines you can use the **parentNode**  method, which will select the container that your current target is inside of. You can chain a couple of these together to take a few steps outside of your target, and then use **remove()** to remove the whole container:

`const target = event.currentTarget.parentNode.parentNode;`
`target.remove()`

**2. Create Product**
Your logic looks great for this one, but instead of creating the rowHtml string and then += each element step by step, you can copy and paste that whole block of code in the innerHTML of the selected element: 

  `table.innerHTML = '<tr class="product">`
 ` <td class="name">`
  `<span>${nameOfNewProduct}</span>`
  `</td>`
  `<td class="price">$<span>${priceOfNewProduct}</span></td>`
  `<td class="quantity">`
   ` <input type="number" value="0" min="0" placeholder="Quantity" />`
  `</td>`
  `<td class="subtotal">$<span>0</span></td>`
  `<td class="action">`
    `<button class="btn btn-remove">Remove</button>`
  `</td>'`;

PR #2487, opened 2022-04-27
Solved lab Cart

ihab5520:


js/index.js
```diff
@@ -1,42 +1,107 @@
...
+    <td class="action">
+      <button class="btn btn-remove">Remove</button>
+    </td>
+  </tr>
+`;
+cart.appendChild(newProduct); 
+nameElement.value = ""; 
+priceElement.value = ""; 
+
+var deleteButton = newProduct.querySelector(".btn-remove");
```
Matteo-Ironhack, line 91: you should not need to use `var` here; `let` or `const` should work just fine.

js/index.js
```diff
@@ -1,42 +1,107 @@
...
+    '.create-product input[type="number"]'
+  );
+
+  let cart = document.querySelector("tbody"); 
+
+  let newProduct = document.createElement("tr"); 
+  newProduct.className = "product"; 
+
+  newProduct.innerHTML = `
+  <tr class="product">
```
Matteo-Ironhack, line 73: no need to create a new `tr` inside the already existing one. check what innerHTML does.
used it like this, it will create
```
<tr>
  <tr>
    ...other elements here
  </tr>
</tr>
```

PR #2488, opened 2022-04-27
horror!!!!

EvgeniiaRiaguzova:


js/index.js
```diff
@@ -1,24 +1,36 @@
 // ITERATION 1
-
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
 
-  //... your code goes here
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector('.quantity span');
```
Matteo-Ironhack, line 6: the quantity is not defined as a span element, but as an input.
one way to get its value could be `product.querySelector(".quantity").getElementsByTagName("input")[0].value`.

one thing to remember is that getElementsByTagName returns an array (even if there is only one element)

js/index.js
```diff
@@ -1,24 +1,36 @@
...
+  //... your code goes here
 function calculateAll() {
   // code in the following two lines is added just for testing purposes.
   // it runs when only iteration 1 is completed. at later point, it can be removed.
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
+  //const singleProduct = document.querySelector('.product');
+  //updateSubtotal(singleProduct);
   // end of test
+ let product = document.getElementsByClassName('product');
```
Matteo-Ironhack, line 23: getElementsByClassName is also returning an array, A way of solving is using a for loop

PR #2494, opened 2022-04-28
domcartlab

zarahsara:


js/index.js
```diff
@@ -14,10 +22,17 @@ function calculateAll() {
   // end of test
 
   // ITERATION 2
-  //... your code goes here
+  const Product = getElementsByClassName('product');
+}
+let i = 0;
+for (let i =0; i < product; i ++) {
```
Matteo-Ironhack, line 28: you might have wanted to loop over 	`Product`, not `product`. Be careful using capital letters and remember that Javascript is case sensitive (meaning whether a letter is capitalized or not makes a difference when naming things)

PR #2535, opened 2022-06-07
Joao-Berlin-FTWD

FullerJoao:


js/index.js
```diff
@@ -15,9 +20,24 @@ function calculateAll() {
...
+  })
 
   // ITERATION 3
   //... your code goes here
+  let allSubtotals=document.querySelectorAll(".subtotal span")
+  console.log(allSubtotals)
+
+  let totalPrice=0
+
+  allSubtotals.forEach((subtotal)=>{
```
mloelkes, line 35: If you would return the subtotal value from the updateSubtotal-function, you don't need this second loop

js/index.js
```diff
@@ -26,17 +46,53 @@ function removeProduct(event) {
...
   //... your code goes here
+  let removeProductBtn = document.querySelectorAll(".btn-remove");
+
+  removeProductBtn.forEach((removeBtn)=>{
+    removeBtn.addEventListener('click', removeProduct);
+  })
+
+  let createProductBtn = document.getElementById('create')
+  createProductBtn.addEventListener("click", createProduct)
 });
```
mloelkes, line 98: Good work! ⭐⭐⭐

PR #2536, opened 2022-06-07
Solved lab

PedrovictorSouza:


js/index.js
```diff
@@ -1,42 +1,92 @@
...
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
-  // end of test
-
-  // ITERATION 2
-  //... your code goes here
-
-  // ITERATION 3
-  //... your code goes here
+  const singleProduct = document.querySelectorAll('.product');
```
mloelkes, line 21: A better name would be allProducts

js/index.js
```diff
@@ -1,42 +1,92 @@
...
+  const allSubTotal = document.querySelectorAll('.subtotal span');
+  let total = 0;
+  let result = document.querySelector('#total-value span');
+  
+  for (let i = 0; i < singleProduct.length; i++) {
+    updateSubtotal(singleProduct[i]);
+    
+    let num = parseInt(allSubTotal[i].innerHTML);
```
mloelkes, line 29: You could also return the subtotal price from the function "updateSubtotal" instead of getting the value from the document again

js/index.js
```diff
@@ -1,42 +1,92 @@
...
 }
 
 // ITERATION 5
 
 function createProduct() {
   //... your code goes here
+  const input = document.querySelector('.create-product input');
+  const price = document.querySelector('.create-product :nth-child(2)');
+  let tb = document.querySelector('tbody');
+  let newRow = tb.insertRow();
```
mloelkes, line 54: You could also insert the whole html element as one string

js/index.js
```diff
@@ -1,42 +1,92 @@
...
+
+  //newRow.cells.innerHTML = input.value;
+  //newRow.classList.add('name');
+  //newRow.ins = newRow.insertCell();
+  console.log(price.value);
 }
 
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
+  const removeItem = document.querySelectorAll('.btn-remove');
```
mloelkes, line 80: A better name would be removeItemButtons

js/index.js
```diff
@@ -1,42 +1,92 @@
...
+  createNewProduct.addEventListener('click', createProduct)
   calculatePricesBtn.addEventListener('click', calculateAll);
+  //console.log(removeItem);
+
+  for(let i = 0; i < removeItem.length; i++) {
+    removeItem[i].addEventListener('click', removeProduct);
+  }
 
   //... your code goes here
 });
```
mloelkes, line 92: Good work! ⭐⭐⭐

PR #2537, opened 2022-06-07
solved lab

YuruBeDeniz:


js/index.js
```diff
@@ -1,42 +1,84 @@
...
-  updateSubtotal(singleProduct);
+/*   const singleProduct = document.querySelector('.product');
+  updateSubtotal(singleProduct); */
   // end of test
 
   // ITERATION 2
   //... your code goes here
-
+  let products = document.querySelectorAll('.product')
+  for(let i = 0; i<products.length; i++){
```
mloelkes, line 34: You don't need the first loop, because you update the subtotal value also in the second loop

js/index.js
```diff
@@ -1,42 +1,84 @@
...
 // ITERATION 4
 
 function removeProduct(event) {
-  const target = event.currentTarget;
-  console.log('The target in remove is:', target);
+/*   const target = event.currentTarget;
+  console.log('The target in remove is:', target); */
   //... your code goes here
+  let product = document.querySelector('.product')
+  product.parentNode.removeChild(product)
```
mloelkes, line 60: Here you always remove the first product. If you would use event.currentTarget, you could remove the selected product

js/index.js
```diff
@@ -1,42 +1,84 @@
...
 }
 
+
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
 
   //... your code goes here
+  //window.addEventListener()
+  let removeButton = document.querySelectorAll('.btn.btn-remove');
```
mloelkes, line 78: This variable should be named removeButtons, because it is a collection of buttons

PR #2539, opened 2022-06-07
Ruiz. Fran

FranEjarque:


js/index.js
```diff
@@ -4,20 +4,37 @@ function updateSubtotal(product) {
...
-  updateSubtotal(singleProduct);
+  // const singleProduct = document.querySelector('.product');
+  // updateSubtotal(singleProduct);
   // end of test
 
   // ITERATION 2
   //... your code goes here
 
+  // I need to itinerate throughout the table and get the name of the product, the price and the quantity
+    for ( product of document.getElementsByClassName('product')) {
```
mloelkes, line 26: You don't need this loop, because the second loop also updates the subtotal and calculates the total price.

js/index.js
```diff
@@ -4,20 +4,37 @@ function updateSubtotal(product) {
...
   //... your code goes here
+   let total = 0;
+   for ( product of document.getElementsByClassName('product')) {
+     total += updateSubtotal(product);
+   }
+   document.querySelector('#total-value > span').innerText = total 
+  
 }
 
 // ITERATION 4
```
mloelkes, line 40: Good work! ⭐⭐⭐

PR #2552, opened 2022-06-09
Shopping Cart - WDFT2022 - Berlin

miguellacorte:


js/index.js
```diff
@@ -1,42 +1,125 @@
...
+
+// let productInputNumber = productInput.getElementsByTagName("input [2]");
+
+
+
 function createProduct() {
   //... your code goes here
+
+
+  let productInput = document.querySelectorAll(".create-product input")[0].value;
```
mloelkes, line 68: You could do this query only once and save it in a variable

js/index.js
```diff
@@ -1,42 +1,125 @@
...
+<td class="quantity">
+  <input type="number" value="0" min="0" placeholder="Quantity" />
+</td>
+<td class="subtotal">$<span>0</span></td>
+<td class="action">
+  <button class="btn btn-remove">Remove</button>
+</td>`
+
+
+let nw1 = NewProduct.querySelectorAll('span')[0]
```
mloelkes, line 87: Good idea! You could also use interpolation to insert the values into the string (`${variable1}`)

js/index.js
```diff
@@ -1,42 +1,125 @@
...
+  }
+
+  let el = document.getElementById("create").addEventListener("click", createProduct);
+
+  console.log(el)
+
+  
+
   //... your code goes here
 });
```
mloelkes, line 125: Good work! ⭐⭐⭐

PR #2557, opened 2022-06-10
Marcel Buchholz Berlin

MAG-91:


js/index.js
```diff
@@ -1,23 +1,48 @@
...
+  // console.log(total)
+
 }
 
 function calculateAll() {
   // code in the following two lines is added just for testing purposes.
   // it runs when only iteration 1 is completed. at later point, it can be removed.
   const singleProduct = document.querySelector('.product');
   updateSubtotal(singleProduct);
+  const allSubs = document.querySelector('.subtotal span')
```
mloelkes, line 25: Good work up to now! Just some code lines missing to update the total value with the value from each product ;) Try to add that if you find the time!

PR #2576, opened 2022-06-22
[RMFT June '22] Lisa Hilterhaus: solve lab

cortinaHs:
Iterations 1-3: group work with Christoph Bochniak - discussed problems together, but mainly individual coding

Feedback request for test "Remove products › should allow removal of multiple existing products" -- this test is failing, but it works fine in the browser

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,7 +18,20 @@
           <th>Action</th>
         </tr>
       </thead>
-      <tbody>
+      <tbody id = "productList">
+        <tr id="template" hidden="hidden">
+          <td class="name">
+            <span></span>
+          </td>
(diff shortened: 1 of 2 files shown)


TA-Remote:
Hey Lisa, great work on the lab!

Your code in javaScript is actually all correct and it took me a moment to figure out what could be wrong. 

My closest guess is because you have a hidden row in your html. Your `parentNode`, has a hidden product - a template. The tests are checking all child nodes of class `.product-list` and is not able to remove it - because it has no class `.product` etc

Try to refactor (rewrite) the code without it and see if you can pass the test. 
Let me know if you still have questions / not working :)
Karina

PR #2586, opened 2022-06-23
Lara Rockenstein WD FT June2022

LMRBLN:


js/index.js
```diff
@@ -1,42 +1,121 @@
...
+    // ---> Approach with clone node 
+
+  let oneRow = document.querySelector('.product');
+  //console.log(oneRow);
+  let clonedNode = oneRow.cloneNode(true);
+  parentOfRowToAdd.appendChild(clonedNode);
+  clonedNode.setAttribute("id", "newProduct1");
+  document.querySelector('#newProduct1 :nth-child(1)').innerHTML=name;
+  document.querySelector('#newProduct1 :nth-child(2)').innerHTML=price;
+
```
TA-Remote, line 87: You're very close here to the solution! 

- Remember, that in html you have input fields to enter product name and price.  -> this is the value you want to use
- After you clone the node , you want to set its innerHtml to that value you read from the input
```
clonedNode.querySelector('.name span').innerHTML = productName.value; 
```
`productName` is variable you have to create beforehand, for example:
```
const productName = document.querySelectorAll('.create-product input')
```

TA-Remote (commented):
Great work Lara! 

I left a clue how you can improve iteration 5.
Also you have one possible solution on student portal.

PR #2598, opened 2022-07-13
[FTLIS072022] - Brenda Lopes

Brenda-Lop:


webLisbonTA:
Great job on all mandatory iterations! You were just missing updating the subtotal on the second product

PR #2610, opened 2022-07-14
FTLI072022

magicmajid2511:


webLisbonTA:
Great job on the 2 first iterations, you were only missing updating the subtotal on the other 2 fields

PR #2611, opened 2022-07-14
[WDFTLISJULY2022] Jason Rosa

JasonRosa:
submit - didn't finish because left comp at school


webLisbonTA:
Great job on all mandatory iterations, you were only missing adding the final subtotal

PR #2652, opened 2022-08-10
WDPT BER Patrick

Patrick-Groot:


Patrick-Groot:
I realized during our lesson Thursday that I misunderstood Iteration 4 (the first bonus one) - I don't remove the whole product, I just remove it from the cart (quantity goes back to 0 and the cost goes down accordingly)

index.html
```diff
@@ -19,7 +19,7 @@ <h1>Ironhack Cart</h1>
         </tr>
       </thead>
       <tbody>
-        <tr class="product">
+        <tr class="product" id="duck">
```
trebeljahr, line 22: this id :joy:

js/index.js
```diff
@@ -1,35 +1,39 @@
...
-  // ITERATION 3
-  //... your code goes here
+function calculateAll() {
+  let priceTotal = 0;
+  const priceTotalInnerText = document.querySelector("#total-value span");
+  const productArray = [...document.getElementsByClassName("product")];
+  for (let i=0; i < productArray.length; i++){
+      updateSubtotal(productArray[i]);
+      priceTotal += updateSubtotal(productArray[i]);
+  }
```
trebeljahr, line 22: the above is perfect for a reduce... and nobody did it :( 
```js
const priceTotal = productArray.reduce((total, product) => total + updateSubtotal(product)), 0);
```

trebeljahr: also with your solution you are calling updateSubtotal twice, wasting precious computing resources completely unnecessarily.

js/index.js
```diff
@@ -1,35 +1,39 @@
...
 
 // ITERATION 4
+const removeButtons = [...document.getElementsByClassName("btn btn-remove")];
+removeButtons.forEach(element => {element.addEventListener("click", removeProduct)});
 
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
-  //... your code goes here
+  target.parentNode.parentNode.querySelector(".quantity input").value = 0;
```
trebeljahr, line 32: I think you should remove the target.parentNode.parentNode - not set it's quantity input value to 0.

trebeljahr: so
```js
event.currentTarget.parentNode.parentNode.remove()
```

trebeljahr (commented):
good work. don't fear using reduce ^^ and maybe give the bonus a try :)

js/index.js
Patrick-Groot: Yeah, I know. I misunderstood the task here and only realized that on Thursday during our class when we were recapping  this iteration

js/index.js
Patrick-Groot: Sorry about not reducing! :D

Patrick-Groot:
> good work. don't fear using reduce ^^ and maybe give the bonus a try :)

Thanks for the feedback! I'll be brave next time and reduce! :D

PR #2653, opened 2022-08-10
[PTBER072022] berndHavekost

hvkst:
that was painful, but done

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,42 +1,73 @@
 // ITERATION 1
-
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  const price = product.querySelector('.price span').innerHTML;
+  const quantity = product.querySelector('.quantity input').value;
+  const subtotalSpan = product.querySelector('.subtotal span');
+  const subtotal = price * quantity;
+  subtotalSpan.innerHTML = subtotal;
+  return subtotal;
 }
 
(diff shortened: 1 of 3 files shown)


js/index.js
```diff
@@ -1,42 +1,73 @@
...
 function calculateAll() {
-  // code in the following two lines is added just for testing purposes.
-  // it runs when only iteration 1 is completed. at later point, it can be removed.
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
-  // end of test
-
   // ITERATION 2
-  //... your code goes here
+  const product = document.querySelectorAll('.product');
```
trebeljahr, line 13: just put the [...] around the right side in line 13.
like so: 
```js
  const product = [...document.querySelectorAll('.product')];

```

js/index.js
```diff
@@ -1,42 +1,73 @@
...
+  let productsArray = [...product];
+
+  let subtotal;
+  let total = 0;
+
+  for (let product of productsArray) {
```
trebeljahr, line 19: this is perfect for reducing.

trebeljahr:
```js
const sum = cart.reduce((agg, product) => agg + updateSubtotal(product)), 0)
```

trebeljahr (commented):
excellent work! a few small things here and there could be still simpler and I left comments :)

PR #2655, opened 2022-08-11
Lab solved

avarap:


js/index.js
```diff
@@ -1,42 +1,76 @@
...
+  const newProd = document.getElementsByClassName('create-product');
+
+  const nameProd = newProd[0].cells[0].childNodes[1];
+  const priceProd = newProd[0].cells[1].childNodes[1];
+
+  console.log(nameProd.value);
+  console.log(priceProd.value);
+
+  const cartList = document.getElementById('cart');
+  row = cartList.insertRow(cartList.length - 1);
```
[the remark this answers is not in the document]
trebeljahr: instead try to get the tbody -> like so:
```js
const cartList = document.querySelector('#cart > tbody');
``` 

also this needs to have more elements - remove button etc. 
so row.insertCell(2) and row.insertCell(3) need to be done, however this whole idea can be simplified when using innerHTML... 

I also changed the accessors for nameProd and priceProd to use a better query selector then your triple array index access...

```js 
function createProduct() {
  const nameProd = document.querySelector('[placeholder="Product Name"]');
  const priceProd = document.querySelector('[placeholder="Product Price"]');

  const cartList = document.querySelector('#cart > tbody');
  const price = parseFloat(priceProd.value);

  const newRow = document.createElement('tr');
  newRow.classList.add('product');

  newRow.innerHTML = `
  <td class="name"><span>${nameProd.value}<span></td>
  <td class="price">$<span>${price}<span></td>
  <td class="quantity">
  </td>
  <td class="subtotal">$<span>0<span></td>
  <td class="action"><button class="btn btn-remove">Remove</button></td>
  `;

  newRow.querySelector('button').addEventListener('click', (event) => {
    event.preventDefault();
    newRow.remove();
  });

  cartList.appendChild(newRow);

  nameProd.value = '';
  priceProd.value = 0;
}
```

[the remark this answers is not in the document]
trebeljahr: but this is the super ultra bonus iteration so yeah, don't worry, it's supposed to be quite hard.

js/index.js
```diff
@@ -1,42 +1,76 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  //console.log('Calculating subtotal, yey!');
+  const price = product.querySelector('.price span');
+  const quantity = product.querySelector('.quantity input');
+  let calc = price.innerText * quantity.value;
```
trebeljahr, line 7: I would be careful here with using quantity.value like so. because it might be a string and is converted nicely. putting a parseFloat() makes sure that you don't get a NaN when it can't be converted. I.e. when quantity.value would be say $25, because in the newly created products you have put the $ sign into the span elements and not out of it
```js
`<span>$25</span>`
```
this would break.

js/index.js
```diff
@@ -1,42 +1,76 @@
...
 
 // ITERATION 4
 
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
-  //... your code goes here
+  //target.parentNode.parentNode.parentNode.removeChild(target.parentNode.parentNode);
+  //after trial and error i took parent of parent
+  const parentElement = target.parentNode.parentNode;
```
trebeljahr, line 36: hmmm. why not just delete the row element itself? 
i.e. 
```js
  event.currentTarget.parentNode.parentNode.remove();
}
```

js/index.js
```diff
@@ -1,42 +1,76 @@
...
-  updateSubtotal(singleProduct);
+  //const singleProduct = document.querySelector('.product');
+  //updateSubtotal(singleProduct);
   // end of test
+  let sum = 0;
 
   // ITERATION 2
-  //... your code goes here
+  const cart = [...document.getElementsByClassName('product')];
+  cart.forEach((product) => (sum += updateSubtotal(product)));
```
trebeljahr, line 23: use reduce!!!

trebeljahr: you want a sum from an array -> reduce!!!
```js
const sum = cart.reduce((agg, product) => agg + updateSubtotal(product)), 0)
```

trebeljahr: also the name cart is misleading, since it's not a cart, but an array of products as Html Elements. 

you don't even need it, because you don't re-use cart so you could just write 
`[...document.getElementsByClassName('product')].reduce ...`

trebeljahr (commented):
I put a solution + some comments regarding the bonus iteration and a few things here and there, the rest looks good and works like intended as well :+1:

js/index.js
```diff
@@ -1,42 +1,76 @@
...
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
 
-  //... your code goes here
+  const createBtn = document.getElementById('create');
+  createBtn.addEventListener('click', createProduct);
+
+  const removeBtn = document.getElementsByClassName('btn-remove');
+  for (let i = 0; i < removeBtn.length; i++) {
```
trebeljahr, line 73: forEach

PR #2656, opened 2022-08-12
Solved lab

herrFernandez:


js/index.js
```diff
@@ -1,33 +1,44 @@
...
+
+function calculateAll() {
   // ITERATION 2
-  //... your code goes here
+  const productList = document.querySelectorAll('.product');   // get every tr.product
+  let totalPrice = 0;
+    productList.forEach(function (singleProduct) {             //  call the function updateSubtotal with every tr.product DOM node
+    const subtotal = updateSubtotal(singleProduct);
+    totalPrice += subtotal;
+ });
```
trebeljahr, line 23: the above is much much nicer with reduce. 

```js
const totalPrice = cart.reduce((total, product) => total + updateSubtotal(product)), 0)
```

trebeljahr (commented):
no bonus, but the iterations that are there are done really nicely :) 
give the bonus a try, maybe have a look at Bernds or Antonios PRs I left some comments there regarding the bonus.

PR #2658, opened 2022-08-13
Iterations 1, 2 and 3 solved

JoaoSerafim3001:


js/index.js
```diff
@@ -4,20 +4,35 @@ function updateSubtotal(product) {
...
+  // const singleProduct = document.querySelector('.product');
+  // updateSubtotal(singleProduct);
   // end of test
 
+
   // ITERATION 2
   //... your code goes here
+  let allProductsSum = 0
+  const productsArr = [...document.getElementsByClassName('product')]
+  productsArr.forEach((product) => allProductsSum += updateSubtotal(product))
```
trebeljahr, line 27: this can be better done with reduce. 
```js 
const allProductsSum = productsArr.reduce((sum, product) => sum + updateSubtotal(product)), 0)

trebeljahr (commented):
the things that are already there look good. Try to add the remove button iteration as well :) 

and don't fear using reduce haha :)

PR #2747, opened 2022-10-11
[LISFT202210] Diana Pirlici

DianaMariaa98:


webLisbonTA:
Great job on the lab Diana! Try doing the bonuses next time!
On the first iteration, you could've declared another variable to calculate the total price * quantity (totalTotal) and returned it.

`const price = product.querySelector('.price span'); let quantity = product.querySelector('.quantity input').value; let priceValue = price.innerHTML let subTotal = product.querySelector('.subtotal span'); totalTotal = priceValue * quantity subTotal.innerHTML = totalTotal return totalTotal`

PR #2867, opened 2022-11-30
[FTRM112022] Julia Chernik

juliachernik:


juliachernik:
Tried to solve bonus iteration 4 but was not able to figure out child/parent relationship

PR #2870, opened 2022-11-30
[FTRMT112022] Jonnathan 

Jogopin:


Jogopin:
it wasn't easy but I got it working, there are 2 solutions, one works a little funky ( its commented).
 there are a lot of things that I don't fully understand : /
the interface is working but I'm failing some tests.

PR #2961, opened 2023-01-24
[LISBFT0123] Tomás Belmar 

TomBelmar:


src/index.js
```diff
@@ -1,31 +1,38 @@
...
 
 function removeProduct(event) {
-  const target = event.currentTarget;
-  console.log('The target in remove is:', target);
-  //... your code goes here
+  const target1 = event.currentTarget;
+  console.log('The target in remove is:', target1);
+
+  
+const target2 = event.currentTarget;
```
lzaquine, line 33: Another way to do this is using the parentNode property, like:

`target.parentNode.parentNode.remove();`

This way, you're targeting the `tr` with `class = product` and then using the remove method to remove it!

Another thing you could've added is the calculateAll function to calculate all the prices after you remove one product.

lzaquine (commented):
Great job on the lab, Tomas!

PR #2966, opened 2023-01-24
[LISFT01234] Vanessa Vieira

vanessav22:


src/index.js
```diff
@@ -1,42 +1,60 @@
...
 
-  // ITERATION 3
-  //... your code goes here
+  document.querySelector('#total-value span').innerHTML = emptyValue
+  
 }
 
 // ITERATION 4
 
 function removeProduct(event) {
```
lzaquine, line 30: Another way to do this is to use the method remove(), like:

`target.parentNode.parentNode.remove();`

And to add, you could've passed the function calculateAll() at the end of it to calculate the prices of the remaining products.

src/index.js
```diff
@@ -1,42 +1,60 @@
...
+let parentRemoveLine = target.parentNode.parentNode.parentNode;
+
+parentRemoveLine.removeChild(removeLine)
+  
+
 }
 
 // ITERATION 5
 
 function createProduct() {
```
lzaquine, line 44: Your buttons aren't working because you have the window.addEventListener inside of this createProduct function.
To add the functionality to the Create Product button, you should've done the logic inside of the window.addEventListener just like the other buttons.

```
const addNewLine = document.getElementById("create");
addNewLine.addEventListener("click", createProduct);
```

lzaquine (commented):
Great job on the lab, Vanessa!

PR #2967, opened 2023-01-24
[LISFT0123] Joana Gomes

joanadgomes:


src/index.js
```diff
@@ -1,42 +1,73 @@
...
+  target.parentNode.parentNode.remove()
+
 }
 
 // ITERATION 5
 
-function createProduct() {
-  //... your code goes here
+function createProduct(event) {
+  const productParent= document.querySelector("#parent");
```
webLisbonTA, line 51: Here we need to target the inputs and store their values, target the table, create a new element and try to pass the values from the inputs to the new created element.
Here's a solution you could've used:
```
function createProduct() {

  let newProduct = `
  <tr class="product">
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  </tr>`;

  // Targeting the table

}
```

webLisbonTA (commented):
Great job on the lab Joana!

PR #2968, opened 2023-01-24
[LISFT0123] Mariana Franco

marianamedeirosfranco:


src/index.js
```diff
@@ -1,42 +1,81 @@
...
 
-function createProduct() {
+function createProduct(event) {
   //... your code goes here
-}
 
+  //Add a click event handler to the "Create Product" that will take a function named createProduct as a callback.
+ 
+  //TO BE COMPLETED...
+
```
webLisbonTA, line 60: Here we need to target the inputs and store their values, target the table, create a new element and try to pass the values from the inputs to the new created element.
Here's a solution you could've used:
```
function createProduct() {

  let newProduct = `
  <tr class="product">
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  </tr>`;

  // Targeting the table

}
```

webLisbonTA (commented):
Great job on the lab Mariana!

PR #2969, opened 2023-01-24
[LISFT01234] Paulo Caetano

paulocaetano96:
hell on earth.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,21 +4,44 @@ function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
 
   //... your code goes here
+
+  let price = product.querySelector('.price span');
+  let quantity = product.querySelector('.quantity input');
+
+  
+  let valueNumber = quantity.value;
+  
+
+  let subtotal = valueNumber * price.textContent;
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -4,39 +4,105 @@ function updateSubtotal(product) {
...
+<td class="subtotal">$<span>0</span></td>
+<td class="action">
+  <button class="btn btn-remove">Remove</button>
+</td>`
+
+newProductline.classList.add('product');
+
+let potitioningTR = document.querySelector('tbody');
+potitioningTR.appendChild(newProductline)
+
```
lzaquine, line 86: Your remove button doesn't work for newly added products because when the eventListener runs, the new html wasn't created yet. To fix this, you would need to create the functionality of the button inside of this function.

```
let removeButton = document.querySelectorAll('.btn-remove')
for (let i = 0; i < removeButton.length; i++) {
removeButton[i].addEventListener('click', removeProduct);
}
```
You don't need to change anything that you did before, just add this here as well.

lzaquine (commented):
Great job on the lab, Paulo!

PR #2973, opened 2023-01-24
[LISFT0123] Caio Mazur

caiomazur:


src/index.js
```diff
@@ -1,42 +1,79 @@
...
+    //getRemove[i].onclick = removeProduct(getRemove[i]);
+    let removeChild = target.getRemove.removeChild(getRemove[i]); // Remove the child
+    return removeChild;
+  } */
 }
 
 // ITERATION 5
 
 function createProduct() {
   //... your code goes here
```
joaopdg, line 58: Here you're targeting the inputs and storing their values, targeting the table, creating a new <tr> element and trying to pass the values from the inputs to the new created element.
The logic behind is more or less what you've done, but you're missing some steps:

```
function createProduct() {

  let newProduct = `
  <tr class="product">
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  </tr>`;

  // Targeting the table

}
```

joaopdg (commented):
Great job on the lab Caio!

PR #2978, opened 2023-01-24
[LISFT0123] André Lisboa

andrelisboapt:
solved lab with bonus

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,22 +2,41 @@
 
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  let price = product.querySelector(".price span");
+  let inputNum = product.querySelector(".quantity input");
+  let subTotalPrice = product.querySelector(".subtotal span")
+  
+  return subTotalPrice.innerHTML = price.innerHTML * inputNum.value
+  
 }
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -2,41 +2,113 @@
...
 
+
 window.addEventListener('load', () => {
   const calculatePricesBtn = document.getElementById('calculate');
   calculatePricesBtn.addEventListener('click', calculateAll);
 
+
   //... your code goes here
+const removeButton = document.getElementsByClassName("action")
+for(let i = 0; i < removeButton.length; i++){
```
joaopdg, line 106: When you click the remove button of one of the first two products, instead of removing the product that you clicked, is removing both. 

This is happening because instead of targeting the buttons by the class `.btn-remove` you targeting by the class `.action` so instead of targeting the `button` you are targeting the `td`.

If instead of `document.getElementsByClassName("action")` you use `document.querySelectorAll(".btn-remove")` this problem will be solved:

```
  const removeButton = document.querySelectorAll(".btn-remove");

  for (let i = 0; i < removeButton.length; i++) {
    removeButton[i].addEventListener("click", removeProduct);
  }
```

joaopdg (commented):
Great job on the lab and the bonus André!

PR #2981, opened 2023-01-24
[LISFTWD0123]Farid

Farid-97:


src/index.js
```diff
@@ -2,41 +2,79 @@
...
+let parent=target.parentNode.parentNode.parentNode
+parent.removeChild(remover)
 }
 
-// ITERATION 5
 
 function createProduct() {
-  //... your code goes here
+
+  let text=document.querySelectorAll(".create-product")
```
webLisbonTA, line 38: Just target `create-product` class with the `querySelector(".create-product")`

src/index.js
```diff
@@ -2,41 +2,79 @@
...
+  let textDoc=text.querySelector("input")
+  let productName=textDoc.value
+
+let number=document.querySelector(".create-product")
+let textnumber=number.querySelector("input [type='number']")
```
webLisbonTA, line 43: Need to remove the space in order to target the type number of the input, like 
`let textnumber=number.querySelector("input[type='number']")`

src/index.js
```diff
@@ -2,41 +2,79 @@
...
+let productPrice=textnumber.value
+
+  let productBody=document.createElement(`tr`)
```
webLisbonTA, line 46: `productBody.className='product';` here instead of the `productBody.classList.add('product');` below in order to assign the class name 'product' to the table.

src/index.js
```diff
@@ -2,41 +2,79 @@
...
+   <td>
+     <input type="text" placeholder="${productName}" />
+   </td>
+   <td>
+     <input type="number" min="0" value="0" placeholder"${productPrice}" />
+   </td>
+   <td></td>
+   <td></td>
+   <td>
+     <button id="create" class="btn">Create Product</button>
```
webLisbonTA, line 57: You copied the code of the create product, but you want to add the code of the products, like the `remove` button instead of the `create product` and also the unit price and quantity that is above in the HTML file.

webLisbonTA (commented):
Good job on the lab. Check out the notes for a better understanding.

PR #2985, opened 2023-01-25
[LISFT0123] Erika

erikasoaress:


src/index.js
```diff
@@ -1,42 +1,57 @@
...
+  target.parentNode.parentNode.remove();
+  calculateAll();
 }
 
 // ITERATION 5
+// couldn´t figure out this one yet :/
 
 function createProduct() {
-  //... your code goes here
+  let newProduct = document.createElement(".create-product input");
```
joaopdg, line 40: If you're using `document.createElement()` what you need to pass inside the parentheses is the specific html tag that you want to create so in this case it would be something like `document.createElement('tr')`

src/index.js
```diff
@@ -1,42 +1,57 @@
...
+  console.log(newProduct);
+  newProduct.innerHTML = "T-shirt"
+  let newContent = document.querySelector(".create-product").appendChild(newProduct);
```
joaopdg, line 43: Here you're trying to add the new product to the table but you're targeting the button create product instead of targeting the table it self, so instead of `document.querySelector(".create-product").appendChild(newProduct)` you should do `document.querySelector("tbody").appendChild(newProduct)`

src/index.js
```diff
@@ -1,42 +1,57 @@
...
-  //... your code goes here
+  console.log('The target in remove is:', target.parentNode.parentNode);
+  target.parentNode.parentNode.remove();
+  calculateAll();
 }
 
 // ITERATION 5
+// couldn´t figure out this one yet :/
 
 function createProduct() {
```
joaopdg, line 39: This one was a little bit tricky, I will leave here the solution with comments that explain what every line of code is doing:

```
function createProduct() {

  let newProduct = `
  <tr class="product">
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  </tr>`;

  // Targeting the table

  // Adding the removeProduct function to the remove button of the new product
}
```

joaopdg (commented):
Great job on the lab Erika!

PR #2988, opened 2023-01-25
[LISFT 0123] Gabriel

gbrlcsgn:
I couldn't finish the thrid iteration

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,22 +2,35 @@
 
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
-
+ 
   //... your code goes here
+  const price = product.querySelector('.price span').innerText;
+  const quantity = product.querySelector('.quantity input').value;
+  const subTotal = product.querySelector('.subtotal span');
+  subTotal.innerText = price * quantity;
+
+  return subTotal.innerText;
 }
 
 function calculateAll() {
   // code in the following two lines is added just for testing purposes.
   // it runs when only iteration 1 is completed. at later point, it can be removed.
-  const singleProduct = document.querySelector('.product');
-  updateSubtotal(singleProduct);
+  /*const singleProduct = document.querySelector('.product');
+  updateSubtotal(singleProduct);*/
   // end of test
 
   // ITERATION 2
   //... your code goes here
+  const allProducts = document.getElementsByClassName('product')
+
+  for (let i = 0; i < allProducts.length; i++) {
+    const singleProduct = allProducts[i]
+    updateSubtotal(singleProduct)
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -2,22 +2,35 @@
 
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
-
+ 
   //... your code goes here
+  const price = product.querySelector('.price span').innerText;
+  const quantity = product.querySelector('.quantity input').value;
+  const subTotal = product.querySelector('.subtotal span');
+  subTotal.innerText = price * quantity;
```
webLisbonTA, line 10: Here, you needed to add another variable that would store the value of the `price * quantity` calculation so that you can return that value since we are going to reuse this `updateSubtotal()` function later in our code. 
You could've done it like this:
 ```
const price = product.querySelector('.price span').innerText;
  const quantity = product.querySelector('.quantity input').value;
  const subTotalText = product.querySelector('.subtotal span');
  let subTotal = price * quantity;
  subTotalText.innerHTML = subTotal;
  return subTotal;
```

File: src/index.js [hunk 2]
```diff
@@ -2,22 +2,35 @@
...
-  updateSubtotal(singleProduct);
+  /*const singleProduct = document.querySelector('.product');
+  updateSubtotal(singleProduct);*/
   // end of test
 
   // ITERATION 2
   //... your code goes here
+  const allProducts = document.getElementsByClassName('product')
+
+  for (let i = 0; i < allProducts.length; i++) {
```
webLisbonTA, line 26: Here, you need to create a variable that would store the values that we are getting from our loop.
```
let total = 0;
  for (let i = 0; i < allProducts.length; i++) {
    total += updateSubtotal(allProducts[i]);
  }
```

src/index.js
```diff
@@ -2,22 +2,35 @@
...
+    const singleProduct = allProducts[i]
+    updateSubtotal(singleProduct)
+  }
 
   // ITERATION 3
   //... your code goes here
+  document.getElementsByClassName('#total-value span').innerHTML = allProducts
```
webLisbonTA, line 33: For this iteration, you targeted the `span` that is inside id `total-value`, however, you used the `.getElementsByClassName()`.
You should've use `document.querySelector('#total-value span').innerHTML = total;` 
Also, since we are storing the values that the loop gave us inside the `total` variable, we need to pass it here instead of the HTML element that you were targeting with `allProducts`.

webLisbonTA (commented):
Gabriel, check the comments I left you for a better understanding of this subject.

PR #3052, opened 2023-03-02
[RMT-PT-012023] Wiebke

fraulueneburg-ironhack:
Iteration 1–3 are finished, 4–5 are somewhat finished, but still buggy. For some reason, in the entire HTML document, can I not access the value of the input fields (it will always return 0/null no matter what).

**Some more bugs:**
- remove button does not work on newly created products
- same for calculate prices button (need to find a way to add the newly created items to the existing array)

### DIFF
--- a/index.html
+++ b/index.html
@@ -48,7 +48,7 @@
         <!-- Iteration 2: Add more products here -->
       </tbody>
       <tfoot>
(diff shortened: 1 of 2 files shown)


fraulueneburg-ironhack:
Update: solved iteration 5, fixed those bugs

PR #3068, opened 2023-03-29
Fabian | Solved Lab + Bonus Iterations + Extra Feature

Fabian04HV:
Extra Feature: When the user tries to create a new product without having entered the name in the input, the input receives a red border, indicating that something is wrong.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,16 +1,23 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
+  const price = product.querySelector('.price span')
+  const quantity = product.querySelector(".quantity input")
+  const subtotal = parseInt(price.innerText) * parseInt(quantity.value)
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -1,42 +1,102 @@
...
+  removeButton.addEventListener("click", function(event){
+    removeProduct(event)
+  })
+
+  newProduct.appendChild(nameTd)
+  newProduct.appendChild(priceTd)
+  newProduct.appendChild(quantityTd)
+  newProduct.appendChild(subtotalTd)
+  newProduct.appendChild(removeButtonTd)
+
```
arekiu, line 81: A little tip: you could save a looooot of lines using the method cloneNode 

https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode

arekiu (commented):
Amazing job! The extra feature is welcome too =D

PR #3073, opened 2023-03-29
BER 29.03.2023 Alexandra Strang

bouadloun:


B4n3l1ng (commented):
Really good overall. Don't forget to comment or delete your console.logs, so as to not slow your code down.

PR #3077, opened 2023-03-29
[BER-WDFT-032023] - GERRIT BRINSAT

GBrinsat:
Done

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,40 +3,118 @@
 function updateSubtotal(product) {
   console.log('Calculating subtotal, yey!');
 
-  //... your code goes here
+  const price = product.querySelector(".price span").innerHTML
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -3,40 +3,118 @@
...
+  const target = event.currentTarget.parentNode.parentNode;
   console.log('The target in remove is:', target);
-  //... your code goes here
+
+  target.remove()
 }
 
 // ITERATION 5
 
 function createProduct() {
```
B4n3l1ng, line 55: For this function, instead of slowly adding classes and child nodes to your tablerow, you could perhaps write your HTML structure as a string and store it in a variable, before appending it to your list. Still works, and that's what matters!

B4n3l1ng (commented):
Well done!

PR #3112, opened 2023-04-18
[Lis-wdft-2023] - Lucas Telles

lltelles:


src/index.js
```diff
@@ -2,30 +2,50 @@
...
 
 // ITERATION 4
 
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
-  //... your code goes here
+  console.log('The target in remove is:', target.parentNode.parentNode);
+  target.parentNode.parentNode.remove()
+
```
joaopdg, line 48: Every time you remove a product you need to update the Total Value, to do this you just need to call the function `calculateAll()` at the end of this function:

```
  
  calculateAll()
}
```

joaopdg (commented):
Great job on the lab and the first bonus Lucas, check lab's solution for iteration 5

PR #3118, opened 2023-04-18
[LIS-WD-FT-APR-2023] Margarida Pereira

mapmargaridapereira:


src/index.js
```diff
@@ -1,33 +1,59 @@
...
 }
 
 // ITERATION 4
 
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
+  console.log('The target in remove is:', target.parentNode.parentNode);
+  target.parentNode.parentNode.remove();
   //... your code goes here
```
joaopdg, line 47: Every time you remove a product you need to update the Total Value, to do this you just need to call the function `calculateAll()` at the end of this function:

```
  
  calculateAll()
}
```

joaopdg (commented):
Great job on the lab and the first bonus Margarida, check lab's solution for iteration 5

PR #3119, opened 2023-04-18
LIS-WDFT_APR-2023-Renato Pais  

Renatopais:


src/index.js
```diff
@@ -1,31 +1,54 @@
...
 }
 
 // ITERATION 4
 
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
-  //... your code goes here
+  console.log('The target in remove is:', target.parentNode.parentNode);
+  target.parentNode.parentNode.remove()
```
joaopdg, line 51: Every time you remove a product you need to update the Total Value, to do this you just need to call the function `calculateAll()` at the end of this function:

```
  
  calculateAll()
}
```

joaopdg (commented):
Great job on the lab and the first bonus Renato, check lab's solution for iteration 5

PR #3120, opened 2023-04-18
Joie ----- lab-dom-ironhack-cart-Solved lab

Jogurgithub412:


src/index.js
```diff
@@ -1,42 +1,55 @@
...
-// ITERATION 4
 
+// ITERATION 4
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
-  //... your code goes here
-}
+  console.log('The target in remove is:', target.parentNode.parentNode);
+  target.parentNode.parentNode.remove();
```
joaopdg, line 34: Every time you remove a product you need to update the Total Value, to do this you just need to call the function `calculateAll()` at the end of this function:

```
  
  calculateAll()
}
```

src/index.js
```diff
@@ -1,42 +1,55 @@
...
+
+
+
+  }
+
+  const removeButton = document.getElementsByClassName('btn-remove');
+  for (let i=0; i<removeButton.length; i++){
+    removeButton[i];
+    removeButton[i].addEventListener('click', removeProduct);
```
joaopdg, line 43: This block of code should be inside the window.addEventListener that you have below

joaopdg (commented):
Great job on the lab and the first bonus Joie, check lab's solution for iteration 5

PR #3125, opened 2023-04-18
[LIS-WDFT-APRIL-23] Guilherme Dantas

Gui-Dantas:


src/index.js
```diff
@@ -4,28 +4,51 @@ function updateSubtotal(product) {
...
 }
 
 // ITERATION 4
 
 function removeProduct(event) {
   const target = event.currentTarget;
-  console.log('The target in remove is:', target);
+  console.log('The target in remove is:', target.parentNode.parentNode);
+  target.parentNode.parentNode.remove();
+
```
brunorocha20, line 48: You correctly created the function to remove a product, however, you are not telling your code to "listen" for the click to trigger the action (call the function to remove the determined product).
Just like you have inside `window.addEventListener()` something that triggers the `calculateAll` function when there's a click in an element with the id `calculate`, you needed to add something similar to this, for when the user clicks to remove the product.
Here's a possible solution for you to add inside that event listener:
 ```
const buttonRemove = document.getElementsByClassName("btn-remove");
  for (let i = 0; i < buttonRemove.length; i++) {
    buttonRemove[i].addEventListener("click", removeProduct);
  }
```

brunorocha20 (commented):
Good job on the lab Guilherme.

PR #3127, opened 2023-04-19
 [LIS-FT-APR23] roshan pradhan

roseanprime:


src/index.js
```diff
@@ -1,31 +1,36 @@
...
-  // end of test
-
-  // ITERATION 2
-  //... your code goes here
+  const products = document.getElementsByClassName('product');
 
-  // ITERATION 3
-  //... your code goes here
+  for(let i=0;i < products.length; i++){
+    updateSubtotal(products[i]);
```
joaopdg, line 20: In this iteration you forgot to update the Total value, to do that you need to declare a variable where you will add the prices from all the products, and then you need to target the HTML tag for the total value and update it's value:

```
function calculateAll() {
  const products = document.getElementsByClassName("product");

  let sum = 0;

  for (let i = 0; i < products.length; i++) {
    updateSubtotal(products[i]);
    sum += updateSubtotal(products[i]);
  }

  let totalValue = document.querySelector("#total-value span");
  totalValue.innerHTML = sum;
}
```

joaopdg (commented):
Good job on the lab Roshan, check the correction that I left you below

PR #3130, opened 2023-04-19
[LIS-FT-APR23]NIROJ

Niroj167:


src/index.js
```diff
@@ -2,24 +2,29 @@
...
-  // ITERATION 2
-  //... your code goes here
+  let total = 0;
+  const products = document.getElementsByClassName("product");
 
-  // ITERATION 3
-  //... your code goes here
-}
+  for (let i = 0; i < products.length; i++) {
+updateSubtotal(products[i]);
```
joaopdg, line 24: In this iteration you forgot to update the Total value, to do that you need to declare a variable where you will add the prices from all the products, and then you need to target the HTML tag for the total value and update it's value:

```
function calculateAll() {
  const products = document.getElementsByClassName("product");

  let sum = 0;

  for (let i = 0; i < products.length; i++) {
    updateSubtotal(products[i]);
    sum += updateSubtotal(products[i]);
  }

  let totalValue = document.querySelector("#total-value span");
  totalValue.innerHTML = sum;
}
```

joaopdg (commented):
Good job on the lab Guy, check the correction that I left you below

PR #3142, opened 2023-05-06
[WDPTAPR 23] Andres Franco y Victor Tovar

Andresfrla:


src/index.js
```diff
@@ -2,30 +2,51 @@
...
+  calculateAll()
+}
+
+const removeButton = document.getElementsByClassName('btn-remove');
+for (let i=0; i<removeButton.length; i++){
+  removeButton[i];
+  removeButton[i].addEventListener('click', removeProduct);
 }
 
 // ITERATION 5
```
ta-web-mexpt2, line 52: You should try to catch this last one step just for fun

```
function createProduct(event) {

  newTableRow.innerHTML = `
    <td class="name">
    </td>
    <td class="quantity">
    </td>
    <td class="action">
    </td>
  `;

  // clean the fields
  newProdNameInput.value = '';
  newProdPriceInput.value = 0;
}
```

PR #3144, opened 2023-05-17
[FTRMT052023] cédricRocher-philippeMiossec

hymced:


hymced:
together

hymced:
by myself

in this alternative solution, I use a global var to store the structure of a row based on the first one hardcoded in the html file; this in order to avoid writing the html code in the JS file (which may require to maintain 2 parallel versions...). Is there a better way to do this without polluting the global object? would it be considered ok in a real world scenario?

hymced:
- with a closure and an IIFE to make a private variable
- with an hidden html tag

other ideas from next morning: 
- .cloneNode()
- target.closest("tbody") / target.closest("product")

PR #3150, opened 2023-05-17
[FTRMT052023] JuanLozano

Jualolo:
Did Iteration 1 and 2 by myself but couldn't finish the third one :(

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,6 +33,20 @@
           </td>
         </tr>
         <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Coffee Cup</span>
+          </td>
+          <td class="price">$<span>10.00</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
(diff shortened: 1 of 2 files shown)


TA-Remote:
Hey Juan! 
well done with targeting the elements and injecting the values🙌
you are totally on the right path with iter 2 and 3, it can be confusing... but they are connected:
```
// ITERATION 2 + 3 (just moving a bit your code😉)
  
    const product = document.getElementsByClassName('product');
    // console.log(product);
    const productArray = [...product];
    console.log(productArray);
    let sum = 0; // initialize the sum before the forEach
    productArray.forEach(function(e){
      subtotalValue = updateSubtotal(e) // store each result of the function in a variable so that we can add it to the sum
      sum += subtotalValue
    })
    const totalSum = document.querySelector('#total-value span')
    console.log(totalSum);
    totalSum.innerText   = sum  // inject the value into the element ( let totalSumValue = Number(totalSum.innerText); )
    console.log(totalSumValue);
```
If you have any other question, please let us know
Keep it up with your hard work💪
Alessandra

PR #3160, opened 2023-05-17
[FTRMT052023] Lucas Gavião

Lucas-Gaviao:
Not complete, struggled a bit to apply what we learned today.

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,6 +33,19 @@
           </td>
         </tr>
         <!-- Iteration 2: Add more products here -->
+        <tr class="product">
+          <td class="name">
+            <span>Ironhack Bread</span>
+          </td>
+          <td class="price">$<span>10.00</span></td>
+          <td class="quantity">
+            <input type="number" value="0" min="0" placeholder="Quantity" />
+          </td>
+          <td class="subtotal">$<span>0</span></td>
+          <td class="action">
+            <button class="btn btn-remove">Remove</button>
+          </td>
+        </tr>
       </tbody>
       <tfoot>
         <!-- <tr class="create-product">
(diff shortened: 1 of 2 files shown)


TA-Remote:
Hey Lucas,
well done targeting the element and using the methods to inject the new value🙌
here just one possible way to go for iteration 2 and 3 (as they were related):
```
   // ITERATION 2
  const myProd = document.getElementsByClassName("product"); // we have an Array with all the products <tr>
    console.log(myProd)

  // ITERATION 3
  // we create a variable to store the total 
  let totalSum = 0
    // for each of them we want to update the subtotal on clicking on the button
    // if there are no product the subtotal is always 0, as alreay in HTML no need to check
        for(let i = 0; i < myProd.length; i++){
          // (ITER 2)
          result = updateSubtotal(myProd[i]) // (ITER 3)store it in a variable the result of the function
          // (ITER 3) add the resul in the total
          totalSum += result
          console.log(totalSum, 'total')
        }

  // (ITER 3) target the total element and inject the totalsum value
  const totalSub = document.querySelector("#total-value span")
  totalSub.innerText = totalSum.toFixed(2) // OPTIONAL: we can use toFix to make it a string with max 2 decimal, and avoid wierd behaviors

  // if (myProd.length > 0){
  //   for(let i = 0; i < myProd.length; i++){
  //   const result = updateSubtotal(myProd[i])
  //   console.log(result)
  //   }
  //  const totalSub = document.querySelector("#total-value span")
  // totalSub.innerText()
  // total = result
```
  
  I know a looooot of DOM methods... but keep it up with your hard work💪 
  If you have any doubts, just let us know!
  Alessandra

PR #3162, opened 2023-05-18
solved Dom by Nishtha

nishag-26:


src/index.js
```diff
@@ -1,9 +1,20 @@
 // ITERATION 1
 
 function updateSubtotal(product) {
-  console.log('Calculating subtotal, yey!');
-
-  //... your code goes here
+  console.log('Calculating subtota, yey!');
+  const quantity = product.querySelector('.quantity span');
```
TA-Remote, line 5: Inside quantity we don't have a `span` we actually have an `input`, and get that `value` would be a little bit different

src/index.js
```diff
@@ -1,9 +1,20 @@
...
+const price =  product.querySelector('.price span');
+const contentPrice = Number(price.innerText);
+console.log(typeof contentprice);
```
TA-Remote, line 8: Pay attention to the casing, JS is case sensitivy

src/index.js
```diff
@@ -1,9 +1,20 @@
...
+const contentQuantity = Number(quantity.innerText);
+console.log(typeof contentQuantity);
+const multiplied = contentPrice * contentQuantity;
+const  subtotal=  product.querySelector('.subtotal span');
+ 
+subtotal.innerText =multiplied;
+return multiplied;
+console.log(multiplied);
```
TA-Remote, line 16: That line of code is never reached because is after the `return`

src/index.js
```diff
@@ -14,6 +25,16 @@ function calculateAll() {
   // end of test
 
   // ITERATION 2
+const productElements = documents.querySelectorAll(".product");
```
TA-Remote, line 28: here it should be `document` instead of `documents`

TA-Remote (commented):
- Good work
- Pay attention and check step by step when coding, and when things aren't working take a step make and try to make it work
- Try to apply those changes and check if you have changes on the screen

PR #3253, opened 2024-11-16
Done

tsekamaru:


LloydC (commented):
Excellent work. There's nothing to add, your solution is clean and concise. Well done

